PR #3, opened 2016-09-30
Corrections and Add missing translation (Portuguese.po) Part1

Lippe35:


sdottaka:
Thank you!

Lippe35:
Hi,
you are welcome. I do not know how many parts the translation will be divided, but will gradually there are many things to fix.

Thanks

PR #21, opened 2017-03-15
VS2015.2 - Delete another obsolete Project attribute

GreyMerlin:
VS2015.2 - Delete another obsolete Project attribute

* This removes the use of the **Microsoft.CPP.UpgradeFromVC71.props**
	file from three of the .vcxproj files (this props file wasn't used in other project files).
 * This .props file only sets the pre-processor flag `_VC80_UPGRADE` which
	is never used by this software.
 * `_VC80_UPGRADE` is referenced in one place in MFC's **<afxreg.h>** file,
	however that code is also never invoked, and only generates a
	#define that also is never used in these projects


sdottaka:
Thank you!

PR #22, opened 2017-03-15
Show "(Debug)" in About dialog

GreyMerlin:
Show "(Debug)" in **About** dialog 
* Also allow text of `"x86"` in About dialog to be translated.
* Also show `WIN32`, `UNICODE`, `_DEBUG` in Configuration log
	when appropriate.

**Implementation:**
* Src/ConfigLog.cpp - `GetBuildFlags()` - lines 330-341 - new code to add a few
	more configuration define values to the build information string for the 
	config log.
* Src/MergeApp.cpp - `AboutInfo()` - lines 160-161 - add `"(debug)"` and reformat
	`" x86"` to have `" "` separated so that "x86" can be translated.

### DIFF
--- a/Src/ConfigLog.cpp
+++ b/Src/ConfigLog.cpp
@@ -327,8 +327,18 @@ String CConfigLog::GetBuildFlags() const
 {
 	String flags;
 
-#ifdef WIN64
+#if defined WIN64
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you!

PR #24, opened 2017-03-20
Use proper font for View>Use Default Font menu item

GreyMerlin:
### Symptom: 
Use of the `View > Use Default Font` menu selection always 
sets the Directory Tree window to the 12-point `Courier New` font,
which is not the default font for that window.
	
### Discussion:  
WinMerge has an `Options::Font` structure that maintains an 
in-memory copy of 'actual' and 'default' values for the font of
each of the Directory Tree and File View windows.  This structure
is populated at program initialization by procedure `SetDefaults()`
in file Src/OptionFont.cpp.  The 'default' values are generated by
this procedure, the 'actual' values are loaded from the Registry.
	
The existing code only queries the system for the default fixed-
size font (via the default MIME codepage) and then uses this font
as the 'default' value for the `Options::Font` structure for both
Directory and File windows.  This value is typically `Courier New`.
	
However, WinMerge actually uses the Menu font as the actual Directory
window font if no other font is marked as 'Specified' in the 
Registry.  The default Menu font (for English, since Windows 7) is 
`Segoe UI`.  But this use of `Segoe UI` has never been reflected in the 
`Options::Font` structure, nor in the Registry.

By having the proper information in the 'default' value in the  
`Options::Font `structure, the `View > Use Default Font` menu 
 selection now works correctly.
	
### Implementation: 
- The procedure `SetDefaults()` is modified to additionally capture 
the system-wide Menu font information via `SystemParametersInfo()` 
to be used for the Directory Tree window, while continuing to use the 
codepage MIME font for the File View window.
	
- A "helper" function is added: `InitializeLogFont()`.  It should only 
be used within Src/OptionFont.cpp.  It is documented by comment in 
the Src/OptionFont.h file.
	
- The information in the Registry for the `Options::Font` structure 
now always reflects the values of the in-memory structure, and  
the values being used by the windows themselves.
	

### Incidental changes:
- Numerous added or modified comments in Src/OptionFont.cpp
- Reordering of all Registry related code to be in the canonical 
order implied by the layout of the `LOGFONT `structure itself.
- Explicitly cast the three boolean values (Italic, Underline, 
Strikeout) to boolean to invoke the properly typed `Options::Font`
procedures (`InitOption()`, `SaveOption()`)
- Remove two unnecessary `String()` function references relating to 
font in files Src/DirView.cpp and Src/MergeEditView.cpp

### DIFF
--- a/Src/DirView.cpp
+++ b/Src/DirView.cpp
@@ -380,7 +380,7 @@ void CDirView::OnInitialUpdate()
 #endif
 
 	// Load user-selected font
-	if (GetOptionsMgr()->GetBool(String(OPT_FONT_DIRCMP) + OPT_FONT_USECUSTOM))
+	if (GetOptionsMgr()->GetBool(OPT_FONT_DIRCMP + OPT_FONT_USECUSTOM))
 	{
 		m_font.CreateFontIndirect(&GetMainFrame()->m_lfDir);
 		CWnd::SetFont(&m_font, TRUE);
--- a/Src/MergeEditView.cpp
+++ b/Src/MergeEditView.cpp
@@ -3886,7 +3886,7 @@ void CMergeEditView::ZoomText(short amount)
 
 	if ( amount == 0)
 	{
-		nPointSize = -MulDiv(GetOptionsMgr()->GetInt(String(OPT_FONT_FILECMP) + OPT_FONT_HEIGHT), 72, nLogPixelsY);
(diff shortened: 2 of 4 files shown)


GreyMerlin:
Sorry about the four "Merge ..." commits.  They are harmless and have no impact on the code; they are redundant with your "Merge ..." commits.

sdottaka:
Thank you!
> Sorry about the four "Merge ..." commits. They are harmless and have no impact on the code; they are redundant with your "Merge ..." commits.
There is no problem at all.

GreyMerlin:
Thank You !!

PR #25, opened 2017-03-20
Feature/time stars

GreyMerlin:
### Symptom: 
Date columns use a leading `"* "` to indicate which date
is newest.  This makes the column look ragged and awkward.

### Implementation: 
Add an appropriate `"  "` as leading text for date
columns that do not have the `"* "` appended.

### Comment: 
Clearly the two characters `"* "` occupy slightly more
horizontal space than two `"  "` characters, unless a fixed-
font is used.  Nonetheless, the Date portions of the columns 
almost appear aligned with the default Segoe UI font.  

The star is clearly more visible, which is the whole point of it
being added.

### Historical Comment: 
I've been using an equivalent to this patch
in WinMergeU since 11/2009.  I've never been visually
aware of the slight column mis-alignment caused by
proportional-width fonts.

### A visual of Before ...

sample-before

### A visual of After (i.e. with this modification) ...

sample-after


GreyMerlin:
Yikes!! Once again those pesky four Merge commits show up.

Sorry!!!

Any clue what I'm doing wrong?  Or what I should be doing better?

- Bill

sdottaka:
Merged. Thank you!

GreyMerlin:
Thank You !

PR #26, opened 2017-03-21
Fix Memory Leak with Drop Targets

GreyMerlin:
### Symptom:
* There will be 4 memory leaks (each 80 bytes long) reported at
	program termination for each File-View window that is opened.

### Discussion:
 * This is only noticeable with a `_DEBUG` compilation.
 * This will happen when either Text or Binary files are being
	viewed/edited.
 * Each File-View window establishes four Drop Targets and registers
 	them with Windows.  Each also establishes an 'alternate' Drop
	Target for use with various internal Drag-And-Drop operations.
 * When the edit windows are closed, the Drop Targets are properly
	`Revoke()`'ed and deleted.
 * However: the 'alternate' Drop Targets are not deleted ... MemLeak!

### Implementation:
* procedure `OnDestroy()` in file
	Externals/crystaledit/editlib/ccrystaleditview.cpp now deletes
	the 'alternate' Drop Target before deleting the master one.
* Additionally, there is now an `ASSERT()` in procedure
	`SetAlternateDropTarget()` guarding the assignment of the
	'alternate' to make sure that it was still NULL

### DIFF
--- a/Externals/crystaledit/editlib/ccrystaleditview.cpp
+++ b/Externals/crystaledit/editlib/ccrystaleditview.cpp
@@ -1363,6 +1363,7 @@ DoDragScroll (const CPoint & point)
 void CCrystalEditView::
 SetAlternateDropTarget (IDropTarget *pDropTarget)
 {
+  ASSERT(m_pDropTarget->m_pAlternateDropTarget == NULL);
   m_pDropTarget->m_pAlternateDropTarget = pDropTarget;
 }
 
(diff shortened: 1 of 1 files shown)


GreyMerlin:
AND ... I figured out my problem with the "Merge Commits" in the previous two PRs.  I was basing my new branches on my own local Master when I should have been basing them on your **sdottaka/master** in my local repo.  At least that worked this time !!

sdottaka:
Thank you for finding the bug!

GreyMerlin:
Thank You for the three PR merges !!

PR #31, opened 2017-07-28
Three unrelated simple changes

GreyMerlin:
# Three unrelated, but simple, changes
## Typo corrections to new Go and Rust Copyrights

Trivial correction to the new `go.cpp` and `rust.cpp` copyright notices to improve clarity.

## Add VS Filters to various projects

Somewhere in the past, the `*.vcxproj.filters` files were accidentally dropped from various sub-projects.  These files are recovered (for VS2015 projects) from `*_vs100.vcxproj.filters` files, as appropriate. 

Also, add two additional 'folders' to the Solution Explorer for the **MergeLang** project; add the accessible `*.vbs` files to appropriate Solution folder (used only in the Post-Build Event step).  

Add other relevant `*.vbs` file to Batch Files solution folder

The Visual Studio Version Selector needs to have `*.sln` files encoded as UTF-8-BOM, with a blank first line, in order to properly determine which VS version to use (when multiple side-by-side VS installations exist).  This is also necessary to allow the proper 'shortcut' icon to be displayed.

## Fix properties for WinXP targeting

Somehow, the \<PlatformToolset\> property was set to `v140`; it should be `v140_xp`.  I think this happened because I did not have the correct WinXP-enabled toolset downloaded.

Also, the \<WindowsTargetPlatformVersion\> property was set to Win10 for the `Release|x64` compilations; it should be left default.

### DIFF
--- a/Externals/crystaledit/editlib/go.cpp
+++ b/Externals/crystaledit/editlib/go.cpp
@@ -3,7 +3,7 @@
 //  Version:    1.0.0.0
 //  Created:    23-Jul-2017
 //
-//  Copyright:  Stcherbatchenko Andrei Portions by Takashi Sawanaka
+//  Copyright:  Stcherbatchenko Andrei, portions by Takashi Sawanaka
 //  E-mail:     sdottaka@users.sourceforge.net
 //
 //  Go syntax highlighing definition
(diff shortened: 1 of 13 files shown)


sdottaka:
Thank you!

GreyMerlin:
Thanks !!

Sent from Mail for Windows 10

PR #42, opened 2017-10-05
Compile & Link without Level=3 warnings

GreyMerlin:
# Compile & Link without Level=3 warnings

This collection of commits has the primary goal of making **WinMerge** (and sub-projects: poco/**Foundation**, poco/**Util**, poco/**XML**, and **UnitTests**)  compile and link correctly without generating any Level=3 warnings.  The intent has been to eliminate the actual code situations that generate the warnings, not to simply mask the generation of the warnings (e.g. by disabling more warnings).  A number of other cleanup situations happened along the way.

### 1.  Code simplification in the following areas ...
- Use `#pragma once` in all `*.h` files (except for the many external **poco** files: Foundation, Util, XML)
- Eliminate silly `*.inl` files by merging their content directly into the appropriate `*.h` or `*.cpp` files.
- Eliminate useless `//{{AFX_INSERT_LOCATION}}` code from the last lines of `*.h` files

### 2. Modernize "deprecated" code ...
- Eliminate pre-processor definitions `_CRT_SECURE_NO_WARNINGS`, `_CRT_NONSTDC_NO_WARNINGS`, `_CRT_SECURE_NO_DEPRECATE`, and `_CRT_NONSTDC_NO_WARNINGS`
- Use the modern `_xx_s()` function names for data manipulation security
- I included all external **poco** files in this change category

### 3. Eliminate all `#pragma warning(disable:nnnn)`  directives
- Many of the manually disabled warnings were of historical interest only, they had no effect with modern VS2015 or VS2017 compilers
- **An unfortunate exception:** The standard Microsoft file `<shlobj.h>` generates a `warning C4091: 'typedef ': ignored on left of 'tagGPFIDL_FLAGS' when no variable is declared` when (and only when) the XP-capable platform toolset is used.  So all `#include <shlobj.h>` directives are now surrounded by `#pragma warning ...` jackets to prevent a warning that we have no control over.
- I included all external **poco** files in this change category

### 4. Eliminate 32- versus 64-bit data truncation and signed/unsigned warnings.
- Most standard functions that return data-size information (esp. string length) return values of type `size_t` (`unsigned __int64` on 64-bit and `unsigned integer` on 32-bit platforms).  These changes make a first-pass attempt to modernize the code to adapt to this `size_t` situation.
- Since `size_t` is `unsigned` and most associated code was `int` or `long`, many parameters, return types and loop control variables have been changed to `size_t`.  Where large areas of code could be impacted by this, I have simply cast the `size_t` to an appropriate `int` or `long` (to simply mimic current behavior).
- Many small utility functions are now properly capable of handling data larger than 2GB, _however_ it is not my intent with these changes to generally address expansion of data structures past the 2GB boundary; there will need to be data structure and algorithmic changes to eventually move correctly past that 2GB limit.  Indeed, MFC's `CString` data type is limited to 2GB bytes per data item; that will be an interesting  challenge.
- I have tried to use the modern `static_cast<t>()` in all changes.  Where a cast is necessary within an `ASSERT()`, I have used the older classical cast form.  Likewise, classical casts must be used in `*.c` files.
- In a few cases where I noticed the possibility of large `size_t` values being stored into smaller `int` or `long` variables, I have added  `ASSERT()` statements to trap those situations (in DEBUG builds).  I'm sure that there are many similar situations that I did not catch.

### 5. Miscellaneous comments ...
- A few **poco** functions are not called by **WinMerge**, and are no longer included in the compilation.
- The **poco** classes used by **WinMerge** are many versions behind what is available on GitHub.  The [PocoProject](https://github.com/pocoproject/poco) is at version 1.7.9, WinMerge is using version 1.4.3 from 2012-01-16. Poco seems to have addressed warnings by simply adding more `#pragma warning (disable:nnnn)` directives to prevent their appearance.  It would be interesting to understand the **poco** changes over the last 5+ years to see if any (or all) changes are relevant to **WinMerge**.
- The Microsoft `ws2_32.lib` library provides no functions needed by **WinMerge**, and the linker no longer tries to link with it.
- A few minor bugs are squashed.
- All changes have been compiled and tested with both **VS2015** and **VS2017**


sdottaka:
Thank you! Merged.
But I think the second parameter of *printf_s() function is the number of characters but byte size. So I will modify the second argument.

PR #43, opened 2017-10-12
Some `size_t` should be `ptrdiff_t`, 

GreyMerlin:
Two independent commits ...

### Cleanup: Fixup 1: some `size_t` should be `ptrdiff_t`

 * Some of my recent type changes to `size_t` should have been to `ptrdiff_t` (which is signed).
 * In particular, `str_pos()`, `get_coding()` and `fget_coding()` in  **cs2cs.cpp** use the value `-2` as a return flag.  This was being improperly handled via `size_t`.
 * Finding this change led to other areas (in my recent changes) with similar problems, in particular * `next` and `prev` in the `DirrRangeInfo{}` structure, and `Open[]` and `Close[]` in the `RxMatchRes{}` structure.
 * These changes compile and run correctly with both **VS2015** and **VS2017**.

### Cleanup2: `gtest/*` files into Solution Explorer

 * Recent study of the **UnitTests** project led me to want the base **GoogleTest** files to be more visible 
 * Add the `./External/gtest/*.h` and `*.cc files` into the UnitTests Solution Explorer, with appropriate sub-folders
 * The various `*.cc` files (except for `gtest-all.cc`) are marked with the **'ExcludedFromBuild' = True**  because they are #included and not compiled directly.
 * Changes are for both **VS2015** and **VS2017**.

### DIFF
--- a/Externals/crystaledit/editlib/ccrystaleditview.cpp
+++ b/Externals/crystaledit/editlib/ccrystaleditview.cpp
@@ -2798,8 +2798,6 @@ OnUpdateToolsCharCoding (CCmdUI * pCmdUI)
   pCmdUI->Enable (IsSelection ());
 }
 
-size_t str_pos (LPCTSTR whole, LPCTSTR piece);
-
 void CCrystalEditView::
 OnToolsCharCoding ()
 {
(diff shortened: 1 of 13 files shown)


sdottaka:
Thank you!

PR #44, opened 2017-10-13
Level 4 Warning Cleanup, first attempt

GreyMerlin:
I've set **warning level=4** for the WinMerge and UnitTests projects (but not typically for the three 'poco' projects).  This generates _**LOTS**_ of warnings; many categories are not interesting and I certainly do not recommend that the **Level=4** become an official setting.  But some warning categories do seem important, either functionally or aesthetically.   So, I'll slowly be working on the ones that I think are interesting.  Here is a first batch of commits ...

### Cleanup2: More Build security checking options

 * Add **SDLCheck=true** to all compilations for additional "Security Checking".  See [/sdl (Enable Additional Security Checks)](https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks)
 * Make sure that **BufferSecurityCheck=true**.  It had been set to **false** for the three 'poco' projects; the normal default value is **=true**.
 * Both VS2015 and VS2017 project files are changed.

### Cleanup2: Uninitialized Local Variables

 * L4 warning C4701: "potentially uninitialized local variable used".  This commit eliminates all occurrences from WinMerge and UnitTests projects (but not from the 'poco' projects).

### Cleanup2: Cleanup 'Output Style' declarations

 * Output Style values (selecting how to format patches that are generated by the `Tools|Generate Patch` menu dialog) were defined in two unrelated places (with slightly different names), but need to have the
identical values.
 * Furthermore, five of the eight values are not used by WinMerge.
 * And- the values may get stored into the Registry, so they cannot be accidentally changed in the future.

 * These changes simplify this situation by basing both definitions on the `diff.h` values, while leaving the two name groups both existing.
 * All declarations and references to the five unused values are commented-out

### Cleanup2: use named item for empty DirItem.size

 * Clarifies code and eliminates warning "C4245: conversion from `int `to `Poco::File::FileSize`, signed/unsigned mismatch"

### Cleanup2: C4456 "Decl hides previous local decl"

 * Remove a number of C4456 (level=4) warnings.  Declarations hiding previous local declarations is generally sloppy/dangerous practice.
 * and a few other minor cleanup changes

### Cleanup2: C4456 "Decl hides previous local decl" (continued)

 * Remove a number of C4456 (level=4) warnings.  Declarations hiding previous local declarations is generally sloppy/dangerous practice.
 * This commit now eliminates these warnings from WinMerge (but not from the 'poco' sub-projects).

### DIFF
--- a/Externals/crystaledit/editlib/ccrystaleditview.cpp
+++ b/Externals/crystaledit/editlib/ccrystaleditview.cpp
@@ -745,8 +745,8 @@ OnChar (UINT nChar, UINT nRepCnt, UINT nFlags)
               ptCursorPos = ptSelStart;
               if (IsSelection ())
                 {
-                  CPoint ptSelStart, ptSelEnd;
-                  GetSelection (ptSelStart, ptSelEnd);
+                  CPoint ptSelStart1, ptSelEnd1;
+                  GetSelection (ptSelStart1, ptSelEnd1);
             
                   /*SetAnchor (ptCursorPos);
                   SetSelection (ptCursorPos, ptCursorPos);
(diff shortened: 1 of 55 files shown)


sdottaka:
Thank you!

Issue #48, opened 2017-12-11
Additional Information Regarding VS2017 C++17 vs WinMerge


GreyMerlin:
Thank You for merging my Pull Request #47.  I fear that it is just the first crack in a "Pandora's Box" situation.

I have just today discovered this note ([https://blogs.msdn.microsoft.com/vcblog/2017/12/08/c17-feature-removals-and-deprecations/](https://blogs.msdn.microsoft.com/vcblog/2017/12/08/c17-feature-removals-and-deprecations/) dated 8 December 2017) on MSDN's **Visual C++ Team Blog**.  

It explains in much greater detail how **VS2017** (and beyond) will be approaching the **C++14** and **C++17** standards, especially the features that have been and will be deprecated.  Pay particular attention to comments regarding `std:tr1`, which is the issue addressed in PR #47; note that this is really a **GoogleTest** framework problem which that project will have to solve.

As mentioned in PR #47, I was not able to eliminate the problem by changing the language setting to **C++14**.  I now see that, beginning with **VS2017** version 15.5, the new deprecation warning is forced to be an new error by **C++14** (the default language), in preparation for the eventual elimination of the `std:tr1` feature in **C++17**.  

* So today I made a quick test: I explicitly set the "C++ Language Standard" property to **C++17** for our **UnitTests** project.  Alas .... there are now 1608 new compile errors.  Most, but not all, are in the **GoogleTest** framework.  Many, perhaps half, can be silenced by defining one (or more) of the new `_SILENCE_CXX17_xxx` symbols.  Most of the remaining errors directly relate to the fact that indeed **C++17** no longer supports the `std::tr1` construct (which the **GoogleTest** framework uses extensively).  But there remains five new errors, all in `cregexp_poco.cpp` that all relate to the **C++17** elimination of `std::auto_ptr`

* So I decided to make another quick test.  I set the language for the **Merge** project to **C++17**.  Alas ... there are now 618 new errors!  Most can be silenced by defining one or more of the `_SILENCE_CXX17_xxx` symbols (as with **UnitTests**);  these new silence-able errors are all in the **boost** library!!  There remain 11 errors that cannot be silenced: eight in `cregexp_poco.cpp` and three in `MergeDoc.cpp`.

So ... all of this leaves me with a number of thoughts ...
[...]

PR #49, opened 2017-12-19
Compile Merge with C++17 language rules.

GreyMerlin:
## Compile Merge project with C++17 language rules.

 * Force the **Merge** project to be compiled with **C++17** language semantics.
 * Use `unique_ptr<>` to replace `auto_ptr<>`
 * Ignore future (**C++20**?) deprecation warnings.

 * Force the **UnitTests** project to be compiled with the **C++14** language semantics.  To move this project forward to **C++17** will require a Google Tests framework that is compatible with **C++17**.

See Issue #48 (_**Additional Information Regarding VS2017 C++17 vs WinMerge**_).  This commit addresses some issues, as noted above and in my reply to #48 today (19 Dec '17).  However, Issue #48 should remain open, as there are more issues to be addressed.

### DIFF
--- a/Externals/crystaledit/editlib/cregexp_poco.cpp
+++ b/Externals/crystaledit/editlib/cregexp_poco.cpp
@@ -30,7 +30,7 @@ using Poco::RegularExpression;
 using Poco::UnicodeConverter;
 
 struct _RxNode {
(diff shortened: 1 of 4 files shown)


sdottaka:
Thank you!

PR #50, opened 2017-12-25
Old-style procedures and C++17 for more projects

GreyMerlin:
### Eliminate old-style procedure declarators

 * Old-style procedure declarators appeared as ...
	`myFunc (a, b, c)`
	`   int a, c; char * b;`
	`{ ... }`
 * This style is very-very old **C** language.
 * In modern C and C++, this procedure would be declared as ...
	`myFunc (int a, char * b, int c);`
       `{ ... }`

 * This eliminates level-4 warning C4131: "uses old-style declarator"
 * Eliminate the awkward `PARAMS()` macro, which was used to simulate the
new style on old compilers that lacked the new style.

 * These patches only effect the DiffEngine

### Compile Merge sub-projects with **C++17** language rules.

 * Force the **ShellExtensions**, and the various **Poco** sub-projects
to be compiled with **C++17** language semantics when using  **VS2017**.
 * Use `unique_ptr<>` to replace `auto_ptr<>`
 * Compile **Merge** with **C++17** when using the **Test** configuration.

 * (unrelated) `coretools.h` is not used by **ShellExtensions**

### Note ...
* **WinMerge** continues to compile and run correctly with **VS2015**.

### DIFF
--- a/Externals/poco/Foundation/Foundation.vs2017.vcxproj
+++ b/Externals/poco/Foundation/Foundation.vs2017.vcxproj
@@ -99,6 +99,7 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <ExceptionHandling>Async</ExceptionHandling>
       <SDLCheck>true</SDLCheck>
+      <LanguageStandard>stdcpp17</LanguageStandard>
     </ClCompile>
     <Lib>
       <AdditionalDependencies>iphlpapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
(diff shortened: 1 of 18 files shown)


GreyMerlin:
Oops... There is a typo in the PR note above. Following is correct ....

* Compile **Merge** with **C++14** when using the **Test** configuration; otherwise the same compilation issues occur as for the **UnitTests** project with **C++17** (released earlier).

The code in the commit is correct, now the comment is correct too !

sdottaka:
Thank you!

PR #51, opened 2017-12-27
UnitTests can now compile with C++17

GreyMerlin:
Two (unrelated) commits ...

## UnitTests can now compile with C++17

* See [my recent comment in Issue #48](https://github.com/sdottaka/winmerge-v2/issues/48#issuecomment-354031013).  
* The **GoogleTest**  framework can disable use of the `std::tr1` class entirely!  This is disabled by specifying `GTEST_HAS_TR1_TUPLE=0;` in the Preprocessor Definition compilation property.  The **UnitTests** project does not use TR1 tuples.

 * The deprecation warnings are to note features scheduled to be deimplemented in some future **C++** standard (possibly as early as **C++20**).  They remain valid features in **C++17**.  These warnings (and the resulting errors) can be removed by specifying `_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS;` in the Preprocessor Definition compilation property.  This allows us to watch and study how other projects react to the necessary changes (mostly related to `std::function_name::_Unchecked_Iterators::`).

 * The `GTEST_HAS_TR1_TUPLE=0;` Preprocessor definition is now included in **VS2015** as well, so that both **VS2015** and **VS2017** are both building similar software.

 * All parts of the **WinMerge-v2**, **WinIMerge** and **FreeImage** projects now compile with **VS2017** using the **C++17** language standard.  They also continue to compile with **VS2015** using whatever language standard it uses. Issue #48 can now be closed.

## UnitTests now finds Plugins correctly

 * The x64 configurations of **UnitTests** were failing to load the `DisplayXMLFiles.dll` library, because the test case was looking in the old `c:/program files (x86)/winmerge` directory.  This directory contains an incompatible implementation of the dll file. 
 * This patch allows the test case to now use the `c:/Program Files/WinMerge` directory.

 * All four configurations x64/win32 and Release/Debug now run correctly.

### DIFF
--- a/Src/Merge.vs2015.vcxproj
+++ b/Src/Merge.vs2015.vcxproj
@@ -302,7 +302,7 @@
       <AdditionalOptions>/EHa  %(AdditionalOptions)</AdditionalOptions>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>.;..\Externals\gtest;..\Externals\gtest\include;.\CompareEngines;.\Common;..\Externals\crystaledit\editlib;.\diffutils;.\diffutils\lib;.\diffutils\src;..\Externals\boost;..\Externals\poco\Foundation\include;..\Externals\poco\XML\include;..\Externals\poco\Util\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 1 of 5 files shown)


sdottaka:
Thank you!

PR #59, opened 2018-01-29
Corrections Portuguese

Lippe35:


Lippe35:
Sorry is in conflicts

sdottaka:
Thank you! 
I will change the encoding of the ReadMe-Portuguese.txt from UTF-8 to CP1252 because if the encoding of the file is UTF-8, it will not be displayed correctly when displaying the contents of the file with Installer.

Lippe35:
Yes Its better that way! I make more changes (updates, adds, corrections) is next days in Portuguese translation! 

Thanks

PR #79, opened 2018-03-18
Version issues plus Adapt to VS2017 15.6.1 changes

GreyMerlin:
### Update version as in SetVersion.cmd = 2.15.3 

* Update version (in `.\version.h`) as in `SetVersion.cmd` = 2.15.3

### Adapt to VS2017 version 15.6.1 

* These problems were introduced with the upgrade to **VS2017 15.6.1**.  They still work correctly with version **15.6.2**.
* Once again, the boost library needs to be aware of the newest **VS2017** compiler version, which is now **_MSC_VER == 1913**
* Two functions in `DirView.cpp` and one function in `MovedBlocks.cpp`  were causing a new level=3 warning:   _"warning C4834: discarding return value of function with 'nodiscard' attribute"_.    	This was caused by increased checking on the `std::accumulate()` function.

* Note:: the `isPresent()` function of `MovedBlocks.cpp` is never 	actually called, which masked the obvious bug in the code 	(now repaired).
* These changes also work correctly with **VS2015** (although the new 	checking and warning message are not in **VS2015**)

### Simplify `version.h` file name confusions

- Problems

  * there were two, very different, `version.h` files ...
     - `.\version.h` which simply contains current version defines
     - `.\Src\Common\Version.h` which declares the `CVersionInfo` class  implemented in the `\Src\Common\Version.cpp` file.
   * The files `.\version.h` and `.\Src\res\Merge2.rc` are not visible in  the Solution Explorer tree.

- Solution ...

  * I have changed `.\Src\Common\Version.*` to be   `.\Src\Common\VersionInfo.*`.  This also follows more typical file     naming norms.
  * I have moved these new `VersionInfo.*` files from the **Header Files**    section of the Solution Explorer, into the **MFCGui/Common** section
  * I have added the `.\version.h`file into the **Header Files** section,    so that it becomes a visible part of the project.  I have **not**    modified the name or purpose of this file because it is generated by    `BuildBin.vs2017.cmd` during the release build process.
  * I have added the `.\Src\res\Merge2.rc` to the **Resource Files**    section of the Solution Explorer (so that it becomes visible to the    project).  I have marked that file as "Excluded From Build" (in all    build configurations) because it is directly included from the    `.\Src\res\Merge.rc2`file.

* I have made these changes for both **VS2015** and **VS2017** projects.


sdottaka:
Thank you!

PR #81, opened 2018-03-27
Improve reading of files > 2GB (1)

GreyMerlin:
Now that commit 4e4e46c81427ca34e940b511c31af732ed184e3c has allowed files larger than 2GB (for 64-bit builds), there are a number of places in **WinMerge** that are not prepared to handle files that large.  These will  have to be discovered by carefully focused "trial and error".  These are the first two (of many) problems encountered ...

* This patch repairs two places that `assert`with long file sizes.
* It should be noted that Windows can only read/write in chunks of up to 2GB.  The code around lines 227 to 232 is part of a loop that is capable of reading very large files in sequential 2GB chunks.  It now has the correct setup code to limit the size of each successive read. 
* The `assert` at line 911 is completely unnecessary for both 32- and 64-bit builds.

For further information regarding **files > 2GB**, see [issue #91](https://bitbucket.org/winmerge/winmerge/issues/91/problems-with-files-2gb) in the [Winmerge repo](https://bitbucket.org/winmerge/winmerge/overview) at BitBucket

### DIFF
--- a/Src/diffutils/src/io.c
+++ b/Src/diffutils/src/io.c
@@ -224,10 +224,12 @@ slurp (struct file_data *current)
               current->buffer = xrealloc (current->buffer, current->bufsize);
 #endif /*__MSDOS__*/
             }
-		  assert((current->bufsize - current->buffered_chars) < UINT_MAX);
+          unsigned int bytes_to_read = min((unsigned int)(current->bufsize - current->buffered_chars), INT_MAX);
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you! I've forgotten that the type of the `size` argument of `_read(..., size)` function is not `size_t`(64bit) but `unsigned int`(32bit).

GreyMerlin:
Thanks !!

PR #82, opened 2018-04-02
Two patches related to reading large files (i.e. > 2GB)

GreyMerlin:
These are just the "next" in a series of patches to allow WinMerge to properly handle very large ( i.e. > 2GB) files; although all of the problems noted and fixed here can also be exhibited with files somewhat smaller than that 2GB size.

## Improve reading of files > 2GB (2)

* The main improvement here is to notice "lines" longer than 2^16 chars and forcefully truncate them.  For binary-like text files, long runs of '\x00' and '\xff' characters are tolerated for lines containing up to
2^18 characters.
* This greatly improves the performance of the display and scrolling of "binary-like" files (files whose extension does not match the **Binary File Pattern**).
* Note too that the function `MultiByteToWideChar()` can only handle output lines up to INT_MAX characters (thus, input lines up to INT_MAX/2 characters).  This was the original catalyst for writing this patch.
* I've dropped in a few `// TODO` comments to looping situations that are particularly slow and should be re-designed with very large files in mind.

## Improve reading of files > 2GB (3)

* This is a "kludge" patch to allow for a responsive UI while reading large files.  After every input line is read and processed, `::PeekMessage()` and `->PumpMessage()` are called to allow the UI to process messages.
* A much more robust solution would involve running `CMergeDoc::LoadOneFile()` on threads that are independent of the UI.  This patch could then be easily removed.
* This patch also removes a useless label and `goto`.

## **Important Note** ... 
The reading and display of large files remains seriously slow.  Simple vertical scrolling seems reasonably responsive.  However ... any manipulation of the "shape" or "z-order" of panes within the WinMerge window consumes **seriously** more elapsed time while the Vertical Scroll Bar is being tediously re-calculated (with a completely non-responsive WinMerge window for much of the time).  For this and other reasons ... I don't consider large files to be very useful yet.

### DIFF
--- a/Externals/crystaledit/editlib/ccrystaltextview.cpp
+++ b/Externals/crystaledit/editlib/ccrystaltextview.cpp
@@ -3250,6 +3250,7 @@ int CCrystalTextView::GetSubLineIndex( int nLineIndex )
       m_panSubLineIndexCache->SetAtGrow( 0, 0 );
     }
 
+// TODO: Rethink this, it is very time consuming
   for( int i = m_nLastLineIndexCalculatedSubLineIndex; i < nLineIndex; i++ )
     {
       m_panSubLineIndexCache->SetAtGrow( i, nSubLineCount);
@@ -3607,7 +3608,7 @@ RecalcVertScrollBar (bool bPositionOnly /*= false*/, bool bRedraw /*= true */)
   else
     {
       const int nScreenLines = GetScreenLines();
-      if( nScreenLines >= GetSubLineCount() && m_nTopSubLine > 0 )
+      if( m_nTopSubLine > 0 && nScreenLines >= GetSubLineCount() )
         {
           m_nTopLine = 0;
           Invalidate ();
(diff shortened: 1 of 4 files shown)


GreyMerlin:
Some visuals ...

[Input file ](https://1drv.ms/u/s!AuImETA4Xe7oi4QYGmX_5jMuCndKaw) which I've use to compare to itself.

[Cursor at the end of the first line](https://1drv.ms/u/s!AuImETA4Xe7oi4QaGhyKjHJkfELhxQ)

[Cursor near beginning of last line](https://1drv.ms/u/s!AuImETA4Xe7oi4QVDTBPlCdaeOPn2Q)

sdottaka:
Thank you!

sdottaka:
>This is a "kludge" patch to allow for a responsive UI while reading large files. After every input line is read and processed, ::PeekMessage() and ->PumpMessage() are called to allow the UI to process messages.

@GreyMerlin 
The above patch causes a crash when pressing Alt+Down while reading large files.
I do not know how to fix it, so I will revert only the patch.
Another patch is welcomed.

GreyMerlin:
I’ll be back at the keyboard on 21 May.  This will be a high priority;  actually, I’ll work on a “real” solution to this problem, as well as other issues related to reading very long files

PR #86, opened 2018-04-30
Update Italian.po

Luigi-C:
The correct italian word for the english "Yes" is "Si", where the "i" is not accented.


sdottaka:
Thank you!

PR #87, opened 2018-06-08
Very Long Path and File Names (and more)

GreyMerlin:
## Very Long Path and File Names (and more)

***NOTE:***  Please also see the section ***"(Unrelated) Additional Patches"*** at the end of this document.

### Symptoms

**WinMerge** cannot recurse into deeply nested folder structures, nor can it handle long file names.  In particular, no *path* plus *filename* can be longer than 260 characters.

See [Issue #95 in the WinMerge repo at Bitbucket](https://bitbucket.org/winmerge/winmerge/issues/95/path-longer-than-256-characters-are-marked).

### Discussion

With the Windows NTFS file system, an extended-length path name can be (approx.) 32767 Unicode characters long.  Individual file (and folder) names can each  be 255 characters long.  See  the MSDN article: [Naming Files, Paths, and Namespaces](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396).  

Many, but certainly ***not all***, of the (Unicode specific) file-handling routines in the **WIN32 API** can be convinced to accept extended-length path names; this is done by adding the characters **` \\?\ `** to the front of path names longer than 260 characters.  This capability has been available since before **Windows XP**.

However ... most Windows "Shell" functions do not support extended-length path names.  Indeed, Windows Explorer itself does not handle these long names.  (Windows Explorer, in the "Anniversary Update" of Win10, supposedly has a Group Policy and/or Registry setting to allow extended-length path names; I cannot get either of these settings to work.)

Additionally, many programs (both Microsoft and 3rd party) cannot handle extended-length path names.  In addition to Windows Explorer: Microsoft **Office**, **Paint**, **cmd.exe**, and **Visual Studio**, plus **GitExtensions, GitHub Desktop**, **NotePad++** and **FastStone Image Viewer** (along with many other 3rd party programs) cannot handle these extended-length path names.

Notably: **OneCommander** (an alternative to Windows Explorer), plus Microsoft's **PowerShell**, Win10's **Photos** and **Paint 3D** all **_do_** correctly handle extended-length path names.

### WinMerge

These commits ( "*Very Long Path and File Names*" ) implement the ability for **WinMerge** to handle extended-length path names in most circumstances.  Specifically ...

- recursing into nested folders to arbitrary depths (to the limit of the underlying file system) and correctly discovering files and folders at each nesting level.

- comparison of all file types (i.e. text, binary, image) at all folder depths.

- opening text, hex or image File Comparison windows for files at all folder depths.

- Works with standard Windows NTFS file systems, plus NAS (Network Attached Storage) systems using standard UNC file naming.

These patches build upon the **poco/foundation** file system; based on the current **1.9.1 poco** release; the **WinMerge** capture of **poco** is version **1.4**.  The **poco 1.9.1** implementation is incomplete with respect to extended-length path names, and my additional changes are included here as well.

#### Limitations

- These patches handle many, but not all, of the right-click "popup context menu" items that are available for folders and files.  These capabilities continue to work correctly until path names become too long, then some of them simply stop working.  I claim that these are all problems with the Windows iShell Interface routines.  There may be future "work around" patches for a few of these context menu items; but I suspect that most problems cannot be solved within **WinMerge** itself.

- The initial ***File Open*** dialog of **WinMerge** can ***Browse*** into deep folder structures (at least on Win10) and the user can type (or paste) extended-length path names; the specified path names are correctly used, and the most recent path names are correctly retained in the system Registry. However the `CComboBox` control that implements the ***File Open*** dialog cannot handle individual text items that are longer than 255 characters; so *previously used names are unfortunately truncated*.  This might be solvable in a future version of **WinMerge** via an "owner draw" scheme.

- The **ShellExtension** does not work with deep folder nesting; the context menu selections do not even appear!  This seems to be problem with the Windows Explorer context menu interface routines.

- ***NOTE:*** I have only tested these patches on **Windows 10 X64 Pro** version 1803.  I have tested both the **Win32** and the **X64** compilations produced by both **VS2015** and **VS2017**.

## (Unrelated) Additional Patches

The following six commits fix various unrelated issues I discovered while working on the extended-length path name patches ...

#### Adapt boost to latest VS compiler version

- **VS 2017** Version 15.7.1 (cl compiler version 19.14.26428.1) updated     `_MSC_VER` to the value `1914`.  This happens ocassionally; **WinMerge** is using a reasonably old version of **boost**.

- This patch updates **boost** to interact correctly, without compatibility warnings.

#### Fix problems with `_countf()` in newest VS 15.7.2

- The `vcruntime.h` version of `_countf()` gives syntax errors starting with **VS 15.7.2**, but only with `CCrystalTextView::m_SourceDefs` (which is local to the  `crystaledit/editlib/ccrystaltextview.cpp` file), and only for **X64** compilations (**Win32** is ok, probably because no alignment issues are involved).  I think that this could be related to **C++17** compliance issues.  

- This patch reverts to a 'traditional' definition of `_countf()`, an already existing (optional) part of the CCrystalTextView package.

#### Hide unused `Poco/Foundation` files from Intellisense

- Use `#if 0` ... `#endif` to force various UNIX, VMS, WINCE and non-Unicode WIN32 files to show as non-compiled code.  **Intellisense**  can then make it obvious that all this code need not be looked at nor patched.

- There are more files that could be marked this way, but these are in the File System implementation where I've been working

#### Remove new compilation error with VS2015

- VS2015 must have tightened up a rule for templates.  This always used to compile.

- now compiles correctly with both **VS2015** and **VS2017**.

#### Prevent faults when "mousing over" Plugin Settings

- With a `Debug` build, an `assert()` would occur when mousing over (or clicking on) the Plugin Settings context menu item *if Plugins are disabled*.

- With a `Release` build, a fault could occur in the same conditions.

#### Three Trivial Changes

- Remove the **poco** `_CHANGELOG` file.  It is only for Apple environments, and has been removed from **poco** version 1.9.1 (sometime after version 1.4).

- The other changes are simply tweaks to comments.


sdottaka:
Thank you!

PR #89, opened 2018-06-17
Various, esp. Improve compare+display of last lines in files 

GreyMerlin:
## Improve interrelation with GNU diffutils source

- Remove four `*.c` files never used by diffutils or WinMerge

- Change `Src/CompareEngines/DiffUtils{.c|.h}` to be `./Wrap_DiffUtils{.c|.h}` to remove confusion about the **diffutils** name.  These files are part of **WinMerge**, not **GNU diffutils**.

- Change Solution Explorer to have the entry **GNU diffutils** rather than **DiffEngine**

- Copy solution changes from **VS2017** to **VS2015**

## Adopt performance improvement from GNU diffutils

- include commit [...414dc606](http://git.savannah.gnu.org/cgit/diffutils.git/commit/?id=68b82f6f8419a815cfcf962b3061352d414dc606) from **diffutils** 2016-10-25 (ver 3.5)

- this commit improves the performance of the TOO_EXPENSIVE optimization on "modern" processors (ca. 2016) by upgrading the minimum threshold of this optimization from 255 to 4096.

## Tweak "Fix alignment of the comboboxes in OpenDlg"

- Adjusted by 1 pixel for even better visual alignment

### Symptoms ...

**ASSERT** could occur (with Debug builds), and possibly a fault could occur with Release builds, if the longest file was the Left File and that file did not have an explicit **EOL** marking after the last line of the file, and there were visible differences at the very end of the file.

### Discussion ...

The internal `m_ptBuf[]->m_aLines[]` buffers could end up with different numbers of line entries in situations where the last line of a file had no explicit **EOL** marking.  Normally, the two (or three) `m_aLines[]` buffers will have "ghost" line entries to represent "missing" lines, so that all `m_aLines[]` buffers would be of the same length.  The lack of a final **EOL** could confuse that logic.  At some future points in the comparison logic, the length used for processing the two (or three) buffers is determined simply by the length of `m_ptBuf[0]->m_aLines[]` structure.

### Solution ...

By the end of the `PrimeTextBuffers()` procedure, the two (or three) `m_ptBuf[]->m_aLines[]` buffers will now **always** have the same length.  Furthermore, there will no longer be &lt;null&gt; entries at the end of the `m_aLines[]` buffers.  These will now be deleted (if **all** `m_aLines[]` buffers have them), or they will be replaced with appropriate "ghost" lines (if only **some** of the buffers have final &lt;null&gt; entries)

Also, a number of comments are improved.

### Symptom ...

An **F5** Refresh could produce slightly different comparison results (near the end of the file) when some input files did not have explicit **EOL** marking after the last line of the file.

### Discussion ...

The original File Compare window is generated directly from the original files.  However, subsequent **F5** Refresh displays are generated from temporary _copies_ of the files that are made after they are read for the first time; these temporary files would **_never_** have **EOL** markings following the last line of the file - thus causing slightly different output.

### Solution ...

When making internal temporary copies of the original input files, a final **EOL** marking will be written to the temp file **_iff_** the original file had a final **EOL**.

## Improve "Options > Color > Differences" dialog layout

- Allow more room for the various lines of caption text by sliding the various color selection controls slightly to the right.

- There were two captions that were each folded onto two lines; these now display nicely with each on a single line.


sdottaka:
Thank you!

GreyMerlin:
I have encountered new (!) problems when merging differences (via the "**Copy Right (Alt+Right)**" or "**Copy Left (Alt+Left)**" commands) when those differences are at the very end of an input file, and will be copied onto the very end of another file.  This happens with both 2-way and 3-way comparisons.

I strongly suspect that this is caused by my  very recent commit [**51519fc**](https://github.com/sdottaka/winmerge-v2/pull/89/commits/51519fc848e23494510d8c8268f7a734c1de3e97) "_Improve compare+display of last lines in files (1)_"

I also note various unexpected errors when running the X64 **Test** build, and different errors when running the Win32 **Test** build.  I don't have much experience with the **Test** build, but at least some of these errors _seem_ different with and without the commit I mentioned.

I'm actively investigating and will repair and report ASAP.

Src/MergeDoc.cpp
```diff
@@ -2028,11 +2026,46 @@ m_ptBuf[file]->SetLineFlag(i, LF_INVISIBLE, false, false, false);
...
-	// but now we keep all diffs, including trivial diffs
+	// Buffers `m_ptBuf[]` may have a final line entry that is <null>.
+	// (a) If ALL buffers have that final <null>, remove them all.
+	// (b) If only SOME have that final <null>, change those <null>
+	//		lines into proper ghost-image lines.
+	// Note too: By this point all buffers must have the same number 
+	//		of line entries; eventual buffer processing typically
+	//		uses the line count from `m_ptBuf[0]` for all buffer processing.
 
+	int nFinalNullLines = 0;
```
sdottaka, line 2037: This code block seems to be the cause of #106 problem.
Also, with this code block, EOL on the last line can not be deleted by DEL key etc.
I did not find a good fix for this, so I would like to disable this code.

GreyMerlin:
Interesting... I cannot look at this (or any other) code right now.  But I recall that this was the first of a number of “fixes” for various end-of-file-handling problems.  I also recall that each of the fixes actually did fix problems I could reproduce without my series of patches. So ... Very Interesting ...

Regarding the fix you have here ...  if it works then it works !!

Clearly I have a few things to study when I return early next month.

Thanks for the fix !

Bill

PR #90, opened 2018-06-27
Improve Handling of Last Lines in Files, *Plus More*

GreyMerlin:
# Improve Handling of Last Lines in Files, *Plus More*

This builds upon my previous commits _**"Improve compare+display of last lines in files (n)"**_ (see GitHub [Pull Request #89](https://github.com/sdottaka/winmerge-v2/pull/89)).  In follow-on comments to that PR, I indicated that problems still existed ...

- when merging differences (via the "**Copy Right (Alt+Right)**" or "**Copy Left (Alt+Left)**" commands) when those differences are at the very end of an input file, and will be copied onto the very end of another file.

Of particular note is the new patch  _**"Improve handling of last lines in files (5)"**_ below, which fixes most of those problems (certainly all of the fatal aborts).  As noted below, there remains a problem with **Copy Left** (and **Copy Right**) at the *very end* of the files: there is an extra **`EOL`** inserted at the very end of the destination file (expect more patches!).

The other patches here contain minor code optimizations or textual cleanup.

## Adopt newer GNU diffutils (2001-11-24, ver 2.7.3 ) (1)

- Remove unused files (unused in both **WinMerge** and newest **diffutils**)
- Remove some unused defines and functions

**Note:** This is only the first ***tiny*** step in migrating from GNU diffutils (11/10/1993, **ver 2.5**, still in use by **WinMerge** now) forward to their current version (05/21/2017, **ver 3.6**).  See discussion within  [Issue #98](https://bitbucket.org/winmerge/winmerge/issues/98/upgrade-diffutils-for-better-comparison)

## Improve handling of last lines in files (1)

- Simply tweaks to comments and debug code
- A few tiny optimizations

## Improve handling of last lines in files (2)

Optimization for `CGhostTextBuffer:: RemoveAllGhostLines()`

- Only compress `m_aLines[]` if there actually are "ghost" lines to be discarded

- Only start compression at the first ghost line

## Improve handling of last lines in files (3)

Trivial textual cleanup ...

- Change `ASSERT(0)` to `ASSERT(false)`
- change  `bFirstLineGhost` from `int` to `bool`

## Improve handling of last lines in files (4)

- Slightly improve line and character boundary checking

## Improve handling of last lines in files (5)

- When using **Copy Left** (or **Copy Right**) on a block of line differences at the _very_ end of both files, a fatal abort would occur when copying lines into the shortest of the two end blocks.

- This fatal error would occur when deleting "ghost" lines at the end of that shortest block, because the starting location of the "ghost" lines was improperly calculated.

- There remains a non-fatal bug in that an extra **`EOL`** (and thus an extra blank line) is occasionally(?) added to the end of a Copy (Left or Right) when that action involves the text block at the very end of the file.  Copy (Left or Right) continues to work correctly when the difference region being copied is not at the _absolute_ end of the input files.

- The Undo action (`ctrl-Z`) continues to function correctly for all of these situations.

Irrelevant note: In the first two attempts to fix this problem, I tried to fix the problem within the **CrystalEdit** classes (where the various aborts actually occurred).  This was a mistake because (1) the root problem was not really within **CrystalEdit**, (2) **CrystalEdit** is a substantial, fully-featured, stable text editor and I certainly did not want to "break" it, (3) the attempted patches got to be large and messy, and were breaking more than they were fixing, and finally (4) I learned enough to find the actual bug in the **`CGhostTextBuffer`** class (which is derived from **`CCrystalTextBuffer`**, but is not part of **CrystalEdit** itself)

## Improve handling of last lines in files (6)

- Hex masking constants (`#define` and `enum` values) are now explicitly marked as **`U`** for Unsigned.

## Fix a few comments in procedure declarations

- Fixed (just a few) comments where the procedure **_definition_** specifies an initial value for a parameter, but the **_declaration_** has an incorrect or missing comment regarding these inital values.

## Fix problems with _countf() in newest VS 15.7.2 (2)

- See my earlier GitHub commit (from 05/27/2018) at [*"Fix problems with \_countf() in newest VS 15.7.2"*](https://github.com/GreyMerlin/winmerge-v2/commit/76754ec1f47042b8b857ad5082054033190788b8).

- This is another instance of that same error, but this error message does not actually abort the compilation (???)


sdottaka:
Thank you!

Regarding upgrading diffutils, I am not familiar with licensing, but as WinMerge's license is GPLv2 and the latest diffutils license is GPLv3, upgrading diffutils may cause license violation.
I do not stick to GPLv2, but I do not know how difficult it is to upgrade WinMerge's license.
It might be better to use diff code like Git.

GreyMerlin:
Thank you for pointing out the licensing issue.  That is not a situation that I want to become involved in 😊.  I don’t think we have any problems at this point, but I will check when diffutils moved to GPLv3.  I will also look at the Git Diff code to see how practical it is to use in the WinMerge context.

I will be away from the keyboard for the next 10 day’s, but will have frequent email access

Thank You
Bill

PR #91, opened 2018-07-17
Attempt to fix the issue #100 - Part 2 ... and more

GreyMerlin:
### Attempt to fix the issue #100 - Part 2

- Commit [2f8e009](https://bitbucket.org/winmerge/winmerge/commits/2f8e009c3f473ca0f6a639990d6307ed95d7694e) checks for the existence of the `WinIMergeLib.dll` and abandons the **ImageCompareTest.Open** test if the `.dll` cannot be loaded correctly.  This checking  prevents the inappropriate (and seriously misleading) SEH error later in the code; _**however**_ it allows the test to Pass even when the `.dll` is not present.

    This follow-on patch forces the test to Fail cleanly, with a mildly relevant message, when the `.dll` cannot be loaded correctly.

- Also, the handling of `EXPECT_NE(nullptr, pDoc)` in other tests is improved to prevent SEH errors if `pDoc==nullptr` should ever become `true`.

### Add a new Solution Filter for MergeTest source files

- Add **MergeTest** as a filter folder to the Solution Explorer.

- Move the Files `Test.cpp`, `TestMain.cpp`, `TestMain.h` and `...\gtest\gtest.h` to this new filter folder (they  were in the **MFCGui** folder)

- Add new **MergeTest** solution filter to **VS2015**

### Update project to latest "Platform Version"
 
- With **VS2017** version 15.7.**5** the \<WindowsTargetPlatformVersion\> is now 10.0.**17134**.0

### Additions to .gitignore

- I don't know why these "all of a sudden" became necessary.  The files have always existed and have always seemed to be ignored before ??

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,12 +14,19 @@
 **/Debug
 
 *.exe
+*.dll
 *.pdb
 *.bak
 *.VC.db
 *.VC.VC.opendb
+*.vcxproj.user
+*.ilk
+*.aps
+*.tlb
 
 Testing/Data/ShellFileOpTest/File1.txt
 Translations/WinMerge/MergeLang.rc
 /ShellExtension/dlldata.c
 /ShellExtension/ShellExtension.h
+/ShellExtension/ShellExtension_i.c
+/ShellExtension/ShellExtension_p.c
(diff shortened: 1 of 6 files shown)


sdottaka:
Thank you!

PR #93, opened 2018-08-03
"Ghost Lines" at the end of file comparisons.

GreyMerlin:
# "Ghost Lines" at the end of file comparisons.

Pull Requests [#90](https://github.com/sdottaka/winmerge-v2/pull/90) and [#89](https://github.com/sdottaka/winmerge-v2/pull/89) have solved various problems with the handling of the last lines of a comparison.  PR [#90](https://github.com/sdottaka/winmerge-v2/pull/90) indicated that there were still additional problems to be solved.  I claim that this group of commits ***solves the remainder of those problems***.  As usual, there are also changes to comments, ASSERTs, formatting, etc.

## Improve comments relating to "ghost buffers"

* Correct and extend various parameter comments, especially for **Doxygen** 

* Make sure default parameter values in a function _definition_ are properly commented in the function _declaration_ as well

* Add either `/*virtual base*/` or `/*virtual override*/` comments to various "ghost buffer " related procedure declarations.

* Eliminate (using `#if 0` and `#endif`) the procedures `InternalInsertGhostLine()` and `InsertGhostLine()` which are never used.

## Enable `checkFlagsFromReality()` Debug checking

* The `checkFlagsFromReality()` procedure was never called.  It is now called from two places in `CGhostTextBuffer::RecomputeRealityMapping()`.

* The complete internals of the procedure are now guarded by `_DEBUG`

* The parameter `bool bFlag` is never used and is now eliminated.

## Improve comments relating to "ghost buffers" (2)

* continuing from earlier commit.

### Last Line of files vs "Ghost Lines" (1)

The special case of inserting text into the very last line of a file when that last line is marked as `LF_GHOST`.  Effectively, the new text is supposed to go "before" the Ghost, but mechanically the text is inserted into the Ghost itself, with a new Ghost line appearing at the end of the file.

Later, the Ghost status of both the first and last inserted lines will get straightened out, with the trailing Ghost line becoming a NULL line.  By setting the last line to NULL, the line will eventually be removed or become an actual LF_GHOST line in `CMergeDoc::PrimeTextBuffers()`

### Last Line of Files vs "Ghost Lines" (2)

Handle the special case of `Undo()` at the very end of a file.

`Undo()` checks that the text that is being replaced in the file matches the records being kept in the `m_aUndoBuf[]` structure.  The deletion of text at very end of the file necessary involves removing an additional CRLF (because of an `LF_GHOST` line which was removed too) which causes this Undo comparison-checking to fail.

Following failure, and only if the failure is at the very end of the file, a new secondary check is made to check for the deletion of the additional CRLF.

**NOTE:** I don't particularly like this code because it puts knowledge of "Ghost" lines directly into the code of the `CCrystalTextBuffer` class.  A subsequent patch will factor this into virtual procedures shared with the `CGhostTextBuffer` class.  Spoiler ... (see my following "*... Ghost Lines (4)*" commit)

### Last Line of Files vs "Ghost Lines" (3)

`LineInfo::Delete()` was not correctly handling the case of deleting all text in a line _including_ the CRLF at the end.  This particular situation exists because of deleting text at the very end of a file.

### Last Line of Files vs "Ghost Lines" (4)

Re-implements the Undo handling for the last line (see my final comments in the earlier "*... Ghost Lines (2)*" commit) 

* Move the "Ghost" sensitive parts out of the `CCrystalTextBuffer` class and into the `GhostTextBuffer` class.

* A new virtual procedure `UndoInsert()`now exists in both `CCrystalTextBuffer` and `GhostTextBuffer`, with the WinMerge and Ghost specific checking in the `GhostTextBuffer`class, and the basic deletion code (for undoing an Insert) in the `CCrystalTextBuffer`class.

* Also, modify the `VERIFY()`testing in the Undo paths to correctly interface with the general failure mechanism within `Undo()`

## Reduce Level 4 Build Warnings (12)

* I must have missed these three Level 4 Warning situations in the earlier Pull Request #90
* They were only generated when building the **Test** configuration for the **Win32** platform.

### DIFF
--- a/Src/GhostTextBuffer.cpp
+++ b/Src/GhostTextBuffer.cpp
@@ -715,7 +715,10 @@ void CGhostTextBuffer::RecomputeRealityMapping()
 passingGhosts:
 	ASSERT( i <= nLineCount );
 	if (i == nLineCount)
+	{
+		checkFlagsFromReality();
 		return;
+	}
 	if (GetLineFlags(i) & LF_GHOST)
 	{
 		++i;
@@ -751,7 +754,10 @@ inReality:
 		}
 		m_RealityBlocks.push_back(block);
 		if (i == nLineCount)
+		{
+			checkFlagsFromReality();
 			return;
+		}
 		++i;
 		goto passingGhosts;
 	}
@@ -764,8 +770,9 @@ inReality:
 Check all lines, and ASSERT if reality blocks differ from flags. 
 This means that this only has effect in DEBUG build
 */
-void CGhostTextBuffer::checkFlagsFromReality(bool bFlag) const
+void CGhostTextBuffer::checkFlagsFromReality() const
 {
+#ifdef _DEBUG
 	const int size = static_cast<int>(m_RealityBlocks.size());
 	int i = 0;
 	for (int b = 0 ; b < size ; b ++)
@@ -779,6 +786,7 @@ void CGhostTextBuffer::checkFlagsFromReality(bool bFlag) const
 
 	for ( ; i < GetLineCount() ; i++)
 		ASSERT ((GetLineFlags(i) & LF_GHOST) != 0);
+#endif 
 }
 
 void CGhostTextBuffer::			/* virtual base */
--- a/Src/GhostTextBuffer.cpp
+++ b/Src/GhostTextBuffer.cpp
@@ -853,3 +853,59 @@ GetUndoRecord(int nUndoPos) const
 	return ur;
 }
 
+bool CGhostTextBuffer::		/* virtual override */
+UndoInsert(CCrystalTextView * pSource, CPoint & ptCursorPos, const CPoint apparent_ptStartPos, CPoint const apparent_ptEndPos, const UndoRecord & ur)
+{    
+    // Check that text in the undo buffer matches text in file buffer.  
+	// If not, then rescan() has moved lines and undo fails.
+
+    // we need to put the cursor before the deleted section
+    CString text;
+    const size_t size = m_aLines.size();
+    if ((apparent_ptStartPos.y < static_cast<LONG>(size)) &&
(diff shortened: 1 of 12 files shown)


GreyMerlin:
Moments after I submitted this PR, my installation of **VS2017** decided to update to version [**15.7.6** ](https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-relnotes#15.7.6).

So I recompiled and retested, and "everything" "appears" to be correct.

sdottaka:
Thank you!

Src/GhostTextBuffer.cpp
```diff
@@ -288,6 +290,21 @@ bool CGhostTextBuffer::InsertText (CCrystalTextView * pSource, int nLine,
...
+	if (bSpecialLastLineHandling)
+	{
+		// The special case of inserting text into the very last line of a file when
+		//	that last line is marked as LF_GHOST.  Effectively, the new text is 
+		//	supposed to go "before" the Ghost, but mechanically the text is inserted
+		//	into the Ghost itself, with a new Ghost line appearing at the end of the
+		//	file.  Later (below), the Ghost status of both the first and last inserted 
+		//	lines will get straightened out, with the trailing Ghost line becomming  
+		//	a NULL line.
+		if ((GetLineFlags(nLine) & LF_GHOST) == 0)	// first line still marked GHOST
```
sdottaka, line 302: @GreyMerlin This PR seems to cause the crash of issue #104.
I don't still understand about bSpecialLastLineHandling but if-condition and the comment in line 302 is contradict.
`if ((GetLineFlags(nLine) & LF_GHOST) != 0)`  seems to avoid the crash. Is this fix OK?

GreyMerlin:
Sorry... I am not near PC until next month (December).

But you are correct that if ((...) == 0) doesn’t match comment.  And your proposed != 0 seems to be a proper fix

... Bill

sdottaka:
Thank you for the reply

PR #95, opened 2018-09-05
Improvements to `CSuperComboBox` and Other Changes ...

GreyMerlin:
#### SuperComboBox
As mentioned in earlier comments of mine, previous changes to allow **WinMerge** to handle "**Extended Length File Names**" exhibit various shortcomings based on related (and generally documented) issues in the underlying **Windows** software.  One of these shortcomings is that the **`ComboBoxEx`** controls that handle path names on the "**Select Files or Folders**"  dialog view cannot store or display text strings longer than 255 characters.  Fortunately the Registry interface does not have this problem, so the path name history is correctly preserved in the Registry.

The effect of this main collection of patches "***Improvements to CSuperComboBox***" is to cache (internally to each `CSuperComboBox` instance) the full text of the path names.  Then a stylistically shortened copy of each path name is stored into the `ComboBoxEx` control itself.  The shortened text is presented in the ComboBox control, but the full text is passed into **WinMerge** as the selected file name.  

Thus a long path name (299 characters) like ...

	\\?\C:\Users\Bill\Desktop\DeepDeep\DeepDeepDeep Folder\DeepDeepDeep Folder 2\ 
	     DeepDeepDeep Folder 3\DeepDeepDeep Folder 4\DeepDeepDeep Folder 5\DeepDeepDeep Folder 6\
	     DeepDeepDeep Folder 7\DeepDeepDeep Folder 8\DeepDeepDeep Folder 9\DeepDeepDeep Folder 10\
	     DeepDeepDeep Folder 11\DeepDeepDeep Folder 12\  

... is displayed in the ComboBox  as  (126 characters) ... 

	C:\Users\Bill\Desktop\DeepDeep\DeepDeepDeep Folder\ ... \DeepDeepDeep Folder 10\
	     DeepDeepDeep Folder 11\DeepDeepDeep Folder 12\ 

***Note*** that the **Edit** portion of the `ComboBoxEx` control has always correctly handled arbitrarily long text.

#### And Other Changes ...

There are a number of related and unrelated commits also included in this Pull Request, all individually described at the end of this document ...

. "*Private Build*" is now easy and useful for developers  
.  Update Poco SharedMemory to current standard  
.  "*Open Dialog Auto-Completion*" == "*From MRU list*" is now fixed  
.  Fix "*Spanish translation in Codepage*" error  
.  Add to "*Visual C++ loose*" filter  
.  Improve label text on the *Open Dialog View* tab  

## Improvements to CSuperComboBox

- Trivial changes 

### Improvements to CSuperComboBox (2)

- Comments and simple text edits

### Improvements to CSuperComboBox (3)

- Simplify `IsComboBox()` checking and make it `const` and `private`

### Improvements to CSuperComboBox (4)

- Modernize many `BOOL` to `bool`
- Reorganize layout of `SuperComboBox.h`
- Better internal detection of "*Optional*" third file's control.  Better handling of that control's "empty" state.
- Minor code cleanup, not affecting overall logic.
- Add comments to procedure parameters with default values.

### Improvements to CSuperComboBox (5)

- Don't save OpenDialog state when closed via `Esc` or `Cancel`.  This (the saving of the combobox states) seemed like an odd behavior of `Cancel`, and is not mimicked if the dialog was closed by **WinMerge**'s termination.
- **Note**: This change was completely removed with the following "*Improvements to CSuperComboBox (6)*" commit.

### Improvements to CSuperComboBox (6)

- Continued improvements to `LoadState()` and `SaveState()`

- Completely remove the changes from the "*SuperComboBox (5)*" commit.  Saving the state with the `Esc` key or `Cancel` button really is an important feature.

### Improvements to CSuperComboBox (7)

* **FINALLY** ... the substance of this whole effort !

- Keep a parallel copy of the file name text to allow for extended length path names that are longer than the 255 character limit imposed by `ComboBoxEx`.

- Modify the way that the `CSuperComboBox` is initialized, so that more state information is known throughout the life of the control, not just within the `LoadState` and `SaveState` procedures.

### Improvements to CSuperComboBox (8)

- When inserting 'long' file paths into the ListView associated with the `CSuperComboBox`, replace the middle of the file path string with an ` ... `.

- Since the ListView is limited to 255 characters, and that is itself really too long to view (even in a really wide `SuperComboBox`), the current algorithm breaks long file paths into two 72-character strings (leading and trailing), finds convenient folder name boundaries (if any) thus possibly further shortening those leading and trailing sub-strings and then concatenates them back together with the ellipsis **` ... `** in the middle.

### Improvements to CSuperComboBox (9)

- Eliminate the "AddTemplate" and "AutoAdd" capabilities of `CSuperComboBox`.  They are never used and it is likely that their logic was incomplete or incorrect.
- Eliminate the `bAdd` and `idstrAddText`parameters from the `CSuperComboBox` constructor.  Again, they were never used and have no meaning without the "AutoAdd" capability.

### Improvements to CSuperComboBox (10)

- Proper name for `m_sFullStateText`

### Improvements to CSuperComboBox (11)

- Fix warnings with 32-bit builds

## "*Private Build*" is now easy and useful for developers

- A **Private Build** string can now be `#defined` in `version.h`; the default value is an empty string (as it should be for all released versions of **WinMerge**).   This value propagates into the **VersionInfo** block via the `Merge2.rc` file, and is therefore available to all checking of the **VersionInfo** block of the `.exe` file.

- The **About** dialog is modified to remove the separate line that displayed the **Private Build:** information in the past, as that line overlapped the "Copyright" text immediately below.  The "**Private Build**" info is now displayed on the same line as the **Version**, separated by "**` + `**".  There is room for about 11 characters of **Private Build** text.

- The Configuration Log File now indicates the **Private Build** information following the **Version**, rather that simply stating `(Private Build)`

- Cleanup the mechanics of getting the **Private Build** information into the **About** dialog.

- Also a few trivial code and comment changes.

## Update Poco SharedMemory to current standard
- Eliminate unused parameters

- *(I made these changes long ago when working in this general area of `Poco`.  I never had a good reason to release them (and probably still don't), but they are for textual compatibility with more recent versions of `Poco`, they are harmless and I don't have to worry about them any more !)*

### Update Poco SharedMemory to current standard (2)
- cleanup various `default` parameters and associated comments

## "*Open Dialog Auto-Completion*" == "*From MRU list*" is now fixed

- The option "***Open Dialog Auto-Completion***" (on the "_**Options (General)**_" setting page) failed to work correctly when set to the value "_**From MRU list**_".  This appears to have been true since the **2.13.11+-jp-8** commit on **04/18/2010** when the ComboBox controls for file/folder selection were changed to use the "ComboBoxEx32" common control.  The failure occurs because `FindString()`, which is used in `CSuperComboBox::OnEditchange()`, is only valid for `CComboBox` but not `CComboBoxEx`.

- This patch restores the "**From MRU list**" functionality, but necessarily builds it upon the recent "*Improvements to CSuperComboBox*" changes.

- It should also be noted that "**From MRU list**" provides a very limited functionality when compared to the "**From file system**" value.  This was true even with WinMerge versions **2.13.10** and earlier.

- (*I encountered this problem while testing the `CSuperComboBox` changes.  As mentioned above, this "**From MRU list**" feature has basically no interesting usefulness.  The "**Default**" and "**From file system**" values are (for me) quite useful.*)

## Fix "*Spanish translation in Codepage*" error

- Fix BitBucket [**Issue #108**: "*Spanish translation in Codepage*"](https://bitbucket.org/winmerge/winmerge/issues/108/spanish-translation-in-codepage)
 
## Add to "*Visual C++ loose*" filter

- Add the files `*.idb` and `*.VC.db` to the filter.  These are both large binary files.

## Improve label text on the *Open Dialog View* tab

- Change "**3rd(optional)**" to "**(optional) 3rd**" for visual alignment.

- Improve messages regarding invalid path names by including their ordinals (e.g. **1st**, **2nd**, **3rd**) in the message text.


sdottaka:
Thank you!

PR #97, opened 2018-10-12
Reorganize the OpenView dialog, plus ...

GreyMerlin:
# Reorganize the OpenView dialog, plus ...

# `BOOL` becomes `bool` and related refactoring
### `BOOL` becomes `bool` and related (1)

* Retains `BOOL` where necessary for **MFC** or **Win API** interfaces
* Retains `BOOL` in `*.c` files (GNU **diffutils**, **PCRE** and others) as `bool` is unavailable.
* Mostly retains `BOOL` in **CrystalEdit** and **BCMenu** related files
* Retains `BOOL` in **Poco** and **Boost** source files.
* Retains `BOOL` when used with values other than simply `TRUE` and `FALSE`, such as values of `-1`, or with additional flag bits.  These should be modernized with some future commits.
* Also `TRUE` becomes `true` and `FALSE` becomes `false` where appropriate.

### `BOOL` becomes `bool` and related (2)

* Implicit assignment conversions of `BOOL` to `bool` that were caught only by **VS2015** (as **VS2017** no longer issues warnings in this situation).  However it is good to be explicit.

### _**Issues**_

* **CrystalEdit** and **BCMenu** should be refactored to change `BOOL` to `bool` where appropriate.

# Add icon for "New (3 panes)" menu item

* Courtesy of [**Tim Gerundt**](https://bitbucket.org/gerundt/),  via [BitBucket issue **#118**](https://bitbucket.org/winmerge/winmerge/issues/118/icon-for-new-3-panes)

* I couldn't find a need to use the additional `.bmp`, `.png`, or `.ico` files provided by **Tim**; nonetheless I have captured them locally, but outside of the repository

* ***ISSUE:*** some of the "unused" images could be added to the various `ToolBarEnabled*.bmp` images, thus allowing "**New (3 panes)**" to be directly available from the main toolbar (right next to the existing "**New**" button).

# Reorganize OpenView dialog

### Reorganize OpenView dialog (1)

* Same info, improved layout.  See [BitBucket issue **#117**](https://bitbucket.org/winmerge/winmerge/issues/117/ideas-for-the-open-dialog), including follow-on commentary for a  visual summary of changes.

### Reorganize OpenView dialog (2)

Automatic changes by **VS**'s Resource Editor ...
* It threw away all the comments; I had **git** put them back.
* I kept the other automatic formatting changes.

### Reorganize OpenView dialog (3)

* Improve the **Enable** / **Disable** logic for **OpenView** dialog items that are only valid for **Folder**s or **File**s.

### Reorganize OpenView dialog (4)

* Allow more room for language translation with **Read-Only** and **Swap** buttons.

### Reorganize OpenView dialog (5)

* Update the `English.pot` file and `*.po` files

### *Issues* ...

* The *file-type* icon for each of the three File Path items is not updated when File Paths are Swapped.  This is not a new issue.

* There seems to be a few situations where File Path validation isn't properly handling the 3rd path.  This is not a new issue.

* I wanted to make the **Compare** button be more "obvious" - I had hoped to make the button itself `Green`, with appropriate complementary colored text.  Alas, it seems that the button needs to be "Owner Draw" to do this, so I've back-burner'ed this improvement for now.

* The difference between **Open Project** and **Load Project** on the **Save Project** menu seems to be very confusing (especially since **Load Project** is used nowhere else in the entire **WinMerge** UI).  I would like to understand the difference (if any) and make some improvements to this button.  There is also some funny logic in `UpdateButtonStatesThread()` (in `Src/OpenView.cpp : 730`) regarding **Project** files that should be cleaned up; it ***seems*** like the **OpenView** dialog is partially expecting to be able to handle a `*.WinMerge` Project file in the first **File** selection bar (if the other two **File** selectors are both empty), but then no other logic seems to successfully follow thru with this.  It would be "*really cool*" if the **Compare** button would become an **Open Project** button in this situation.


sdottaka:
Thank you!

PR #100, opened 2018-10-23
Trivial Bugfix: Debugger generates three warnings

GreyMerlin:
**Symptom:** Starting WinMerge with an active Debugger generates three identical warning messages ...
    `... \dlgdata.cpp(254) : AppMsg - Warning: dialog data checkbox value (205) out of range.`

**Solution:** The `m_bReadOnly`array was not being initialized, thus all of the `bool` elements had a value of `0xcd` = 205 when initially given to the three **Read-only** checkboxes on the OpenView dialog.

The array is now properly initialized in the `COpenView::` constructor.

**Also:** An unrelated trivial comment is repaired.


sdottaka:
Thank you!

PR #101, opened 2018-10-28
Various refactorings 

GreyMerlin:
## Various refactorings involving NULL, ASSERT, TRUE and FALSE (1..7)

At this point `NULL`, `BOOL`, `TRUE`, and `FALSE ` have been completely migrated to `nullptr`, `bool`, `true`, and `false`.   Alas, some remain, as they are necessary for (a) Win32 API parameters and results and (b) all `*.c` source files.  I have not modified any of the **Poco Foundation** nor **Google gtest** source files.

Also, related modifications ...
* Many more `NULL` -> `nullptr`
* More `ASSERT (ptr)` -> `ASSERT (ptr!=nullptr`)
* More `if (ptr)` -> `if (ptr != nullptr`)
* Also `if (!ptr)` -> `if (ptr == nullptr)`
* Also, treat `HWND` like `ptr` for these refactorings

* More `TRUE/FALSE` -> `true/false` where `bool` vars already exists
* All `ASSERT(FALSE)` are now `ASSERT(false)`
* Some `if ( i1 & i2 )` -> `if ( (i1 & i2) != 0)`
* Some `b == false` -> `!b`
* Some `b == true` -> `b`

* `assert(0)` -> `assert(false)`
* Some `return 0;` -> `return nullptr;`
* Also `ptr(0)` initializations -> `ptr(nullptr)`

* Change many `return(v);` -> `return v;`
* Many changes `= 0` to `= nullptr`

* The various `*.c` files must use `NULL` instead of `0`

## Cleanup PLUGIN_MODE usage

While modernizing more NULL and BOOL situations, I encountered the following two comments in the
code ...
	`// TODO: Convert to PLUGIN_MODE and fix compile error` and
	`// TODO: Convert bMode to PLUGIN_MODE and fix compile errors`

* Values of type `PLUGIN_MODE` were being stored into `int` variables, often with a `b` prefix (implying the variables were boolean in nature).
* And values themselves were sometimes `false`, or `1` although sometimes from the `PLUGIN_MODE`
enumeration.
* Value testing often assumed the values were boolean (i.e. zero or non-zero)

So ...
* I've changed all appropriate variables to type `PLUGIN_MODE` and used explicit values from the enumeration in all related expressions.
* I've also taken the opportunity to modernize all member variables in the `PluginForFile` and `PackingInfo` classes, by adding an explicit `m_` prefix


sdottaka:
Thank you!
Because it has a large amount of change, it may not be included in 2.16.0. I'm sorry.

Issue #106, opened 2018-11-16
File compare always showing different ending


Antes:
This marks all files different even they are identical. All settings are set to default, no custom regex/white-space command added.

Version: 2.15.5.0 - Unicode x64

https://gyazo.com/d4631094df1b1047d722ab4818812439

sdottaka:
Fixed in 1b52d8491a0c3c0e9bad1daa0fa586c49c9af75d. Thank you for the report.

PR #111, opened 2018-12-08
Remove `nFinalNullLines` checking, disabled by an earlier commit

GreyMerlin:
* Github Commit 1b52d84 "MergeDoc.cpp: Fix github issue #106 ..." effectively removed this code
with a `#if 0`.  This present commit actually removes the erroneous code, but preserves ASSERT checking to
make sure that all `m_ptBuf[]` structures have the exact same length.

* These lines of code (from GitHub Commit 51519fc, 16 Jun 2018) were an early attempt to solve
various problems with the last line(s) of a comparison.  Subsequent changes made these lines
irrelevant.  Independently they were shown to be erroneous.


sdottaka:
Thank you!

PR #118, opened 2019-01-01
OpenDlg: Alternatively to Swap buttons, allow drag & drop between fil…

datadiode:
…etype icons

Unrelated extra payload: Some accelerator keys previously contained in the IDR_MAINFRAME accelerator table have moved to the added IDR_MERGEDOCTYPE accelerator table to prevent them from hiding keystrokes from OpenDlg. This makes COpenView::PreTranslateMessage() dispensable. There still exist more accelerator keys which should follow that move, e.g. VK_F4, whose presence in the IDR_MAINFRAME accelerator table prevents it from opening and closing the list of a focused combobox.


sdottaka:
Thank you!

sdottaka:
To avoid license violation, I replaced the image of rotate2.ico with another image.

PR #119, opened 2019-01-03
OpenDlg: Expose cleaner moving/resizing behavior esp. in non-maximize…

datadiode:
…d state

The proposed change removes the option to move the OpenView window around inside its otherwise empty parent frame, but rather keeps it centered all the time. To retain symmetry, resizing the window by dragging the grip takes effect at both ends.

The changes in MainFrm.* are merely to reduce flicker while the MDI client window messes with its child frames, and can be considered and applied independently.

### DIFF
--- a/Src/MainFrm.cpp
+++ b/Src/MainFrm.cpp
@@ -313,6 +313,8 @@ int CMainFrame::OnCreate(LPCREATESTRUCT lpCreateStruct)
 	if (CMDIFrameWnd::OnCreate(lpCreateStruct) == -1)
 		return -1;
 
+	m_wndMDIClient.SubclassWindow(m_hWndMDIClient);
+
 	m_lfDiff = Options::Font::Load(GetOptionsMgr(), OPT_FONT_FILECMP);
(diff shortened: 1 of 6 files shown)


sdottaka:
Thank you!

GreyMerlin:
While I basically like this change (keeping the OpenView dialog centered), that does lead to the situation where a wide dialog (on a full screen perhaps) can be rendered useless when the WinMerge window is changed to a smaller size.


The only way to recover is to expand the screen again, reduce the size of the OpenView, then reduce the WinMerge window size ... sigh.  Previously, the OpenView dialog could be slid to the extreme left to encounter the corner grabber (also not an "optimal" solution either)..

Trivial for sure, but **perhaps the OpenView dialog size should track the WinMerge screen size when that size shrinks below the width of the dialog**.


Wide Screen
2019-01-03_cr 1

Then a smaller WinMerge window (sorry - the relative sizes don't show up well, but you can still see the effect)
2019-01-03_cr 2

sdottaka:
>Trivial for sure, but perhaps the OpenView dialog size should track the WinMerge screen size when that size shrinks below the width of the dialog.

Fixed in 3e71b0b

PR #120, opened 2019-01-05
Improve Hard-IO error handling, other bug fixes, cleanup, tweaks

GreyMerlin:
### Additional low-level cleanups for ...
     Additional low-level cleanups (2)
     Additional low-level cleanups (3)
     Additional low-level cleanups (4)
     Additional low-level cleanups (5)

* More `BOOL`, `TRUE`, `FALSE` become `bool`, `true`, `false` where appropriate
* Also `while (1)` becomes `while (true)`
* Also `while (ptr)` becomes `while (ptr != nullptr)`
* More use of `ptr != nullptr` in conditionals
* Also `handle!=nullptr`in conditionals
* Also `while (ptr)` becomes `while (ptr != nullptr)`
* Add `{ ... }` brackets to a `do ... while()`
* Some `while (int)` changed to `while (int!=0)`
* More `if (ptr)` changed to `if (ptr != nullptr)`
* More `ptr = 0` changed to `ptr = nullptr`
* More changing `0` parameters to `nullptr` as appropriate
* Cleanup related comments

### Repair Compare Statistics dialogs

* Add proper icon for identical folders
* Account for identical folders in the Folder column
* Properly calculate number of **Binary Identical** and **Binary Different** items
* Examples: both "before" (top window) and "after" (bottom window)

2019-01-04_cr

### Propagate compare FileIO error status to parent directories

* Hard FileIO error status was not being propagated to parent directories, thus showing that the
top-level directories compared as "Same", when in fact there were serious problems deeper in the
folder structure.

* Additionally, if a Folder comparison is aborted with the **STOP** button, the parent directories are
now marked with an error status to make it easier to visually determine which file comparison(s)
were interrupted.

* Example: without patch.  _No errors can be known until the proper folder is expanded._

2018-12-22_cre

* Example: with patch (different, but similar, example).  Notice that the **error** icon is propagated up to the highest folder levels.

2018-12-24 1 _cr

### Distinguish between Aborted and "not yet compared" files

* Add new icon resource `IDI_FILE` (equated currently to the existing "res\\MergeDoc.ico" Icon file)
* Use `IDI_FILE` (rather than `ID_ABORT`) for files that have not been compared
* Continue to use `IDI_ABORT` for files whose comparisons are terminated by the "Stop" button.

### Distinguish between Aborted and "not yet compared" files (2)

* Use a new icon for files "not yet compared".  This new icon is `Src/res/fileunknown.ico`.
Derived from existing icon `Src/res/MergeDoc.ico` with an overlayed blue question mark.

* fileunknown

### Distinguish between Aborted and "not yet compared" files (3)

* Repair visual blemish on icon  

* Example: Using "STOP" during long folder comparison.  Note that 8 threads were actively comparing 8 files when the "STOP" button was pushed.

2018-12-28 2 _cr

### Add `NDEBUG` preprocessor for Release builds

* Some debugging code is still enabled in **Release** builds without `NDEBUG`, even if `_DEBUG` is not specified.  In particular, the `assert()` macro (note all lower case) remains active.
* The Release builds of the various **Poco** libraries have always been compiled with `NDEBUG`,
likewise for the **UnitTests** project.

### Fix obscure memory leak

A memory leak can be caused under the following circumstance:

 * Compare a directory structure with a number of folders and/or files.
 * Select one (or more) items.
 * Outside of **WinMerge** (i.e. without **WinMerge**'s knowledge) delete one (or more) of the files or folders that is already selected in WinMerge.
 * Switch back to **WinMerge**, and then select "View | Refresh Selected" from menu (or type `Ctrl-F5`) to rescan only the selected items (note: some of these are now missing files or folders)

**WinMerge** will correctly detect that the files/folders are missing while rescanning, and will modify its internal directory structure tree accordingly.  HOWEVER - **WinMerge** would **_not_** actually `delete` the disconnected tree item's memory, thus causing a memory leak (perhaps many leaks).  This will be visible with a Debug build of **WinMerge**.

The fix is both simple and allows the surrounding code to be simplified a little bit.

### Propagate compare FileIO error status to parent directories (2)

*  Allow hard FileIO error (or STOP error) status to be propagated to parent directories during `Ctrl-F5` (Refresh Selected) operations.

### Add `hekxedit.h` to project files

* Improve IntelliSense's  handling of `hekxedit.h` declarations
* **VS2015**, **VS2017**, **VS2019**


sdottaka:
Thank you!

PR #122, opened 2019-01-13
OpenDlg: Avoid retaining bogus filetype icons from previous selections

datadiode:
Works by intercepting WM_DRAWITEM sent to ComboBoxEx control, and checking for ODS_COMBOBOXEDIT, to gain control over which icon to display next to the subodinate edit control.

Besides, a possible race condition is addressed which could previously cause a use-after-free situation should the worker thread created when dealing with a network path be delayed beyond the point where its creator deallocates the passed-in string.


sdottaka:
Thank you!

PR #123, opened 2019-01-14
DiffTextBuffer: Create Unicode temp files (makes EscapeControlChars()…

datadiode:
… obsolete)

While transcoding UTF-16 input to a relaxed superset of UTF-8 so diffutils can make sense of it, prepare_text_end() (located in diffutils/src/io.c) translates NUL chars into 2 byte sequences so as to not create bogus string terminators. That was also the single remaining purpose which EscapeControlChars() was still serving after EditLib was enabled to show control characters in hexadecimal notation with a rounded rectangle around them. Now that the NUL char handling built into prepare_text_end() applies, EscapeControlChars() is no longer needed.


sdottaka:
Thank you!

PR #144, opened 2019-05-16
Tweak size of combobox "codepage" in options

gandf:
Size of combobox on "option codepage" not allow to view content.


sdottaka:
Thank you!

PR #145, opened 2019-05-17
Sort combobox codepage and add manual codepage

gandf:
Improve presentation with sort codepage combobox list and add manually codepage in this list if not present.


sdottaka:
Thank you!

PR #156, opened 2019-06-29
coretools.cpp: linelen() should not truncate lines with embedded NULs

datadiode:
The issue with such truncated lines is that line filters are blind for any text to the right of an embedded NUL.


sdottaka:
Thank you!

PR #157, opened 2019-06-29
Favor PathContext::GetSize() over PathContext::size()

datadiode:
As a side effect, this fixes a few warnings and saves a few typecasts which presumably existed to fix a few warnings.


sdottaka:
Thank you!

PR #158, opened 2019-06-30
Consolidate FolderCmp (1)

datadiode:
Wondering why we create additional FolderCmp instances inside FolderCmp::prepAndCompareFiles(), from which we only use the m_diffFileData member, leads me to this proposal to convert them to DiffFileData. This is intended as a first step towards some more invasive rework to reduce code duplication and possibly other sorts of mess.


sdottaka:
Thank you!

PR #160, opened 2019-07-06
Consolidate FolderCmp (2)

datadiode:
Store pointer to CDiffContext as a member in FolderCmp.
Next planned step is to have the folder compare threads create a reusable FolderCmp instance in advance rather than repeatedly create dedicated ones for every single file comparison.


sdottaka:
Thank you!

PR #161, opened 2019-07-06
Consolidate FolderCmp (3)

datadiode:
Have the folder compare threads create a reusable FolderCmp instance in advance rather than repeatedly create dedicated ones for every single file comparison.


sdottaka:
Thank you!

PR #169, opened 2019-08-01
Changed symbols in Newer/Older column for better visual clarity

ex:
I find these symbols more easy to distinguish when comparing folders.
Can help with visual fatigue.


sdottaka:
Thank you!

StoyanDimitrov:
Where exactly those are being displayed?

sdottaka:
Please see the attached image
pr169

PR #183, opened 2019-09-11
Add Japanese manual

gerundt:
This PR adds the Japanese manual from @sdottaka to the official WinMerge. I also tweaked the folder structure from the manual to support different languages of the manuals:

* `Docs\Manual`
  * `EN` *English manual*
  * `JP` *Japanese manual*
  * `Tools` *Place for the build tools*

The only problem is, that I get encoding problems because my Windows has no `Shift_JIS` installed. Maybe it is possible to switch to UTF-8?


sdottaka:
Thank you!
>The only problem is, that I get encoding problems because my Windows has no Shift_JIS installed. Maybe it is possible to switch to UTF-8?

I'll look at it later.

sdottaka:
Unfortunately, "The HTML Help Compiler is not UTF-8 aware" (http://docbook.sourceforge.net/release/xsl/current/doc/html/htmlhelp.encoding.html).
For this reason, I changed to use xsltproc which can use Shift_JIS instead of Saxon as XSLT processor.

PR #190, opened 2019-09-26
Small Manual tweaks

gerundt:
This PR shares XSL and CSS between the Manual translations. For the WinMerge version number we can now use the ENTITY `&WinMergeVersion;`. We have also now ENTITIES for the Manual revision number and copyright informations.


sdottaka:
Thank you!

Issue #200, opened 2019-10-08
[UX] After Comparison the Window is Flashing Indefinitely


Kizaemon:
Please make the window flash count to be 3 or 5, as this is enough to indicate that a lengthy comparison has finished. Currently the window will flash indefinitely, until clicked. This is a distraction, as this forces to interrupt the flow of my current work. 

winmerge/Src/MainFrm.cpp 

Function:
```
/** 
 * @brief Start flashing window if window is inactive.
 */
void CMainFrame::StartFlashing()
{
	CWnd * activeWindow = GetActiveWindow();
	if (activeWindow != this)
---		FlashWindowEx(FLASHW_ALL | FLASHW_TIMERNOFG, 0, 0);
+++		FlashWindowEx(FLASHW_ALL | FLASHW_TIMERNOFG, 3, 0);
}
```

Many thanks!

sdottaka:
Could you make the pull request?

PR #206, opened 2019-10-11
Limit the window flashing count.

Kizaemon:
After a lengthy comparison operation the window is flashing indicating the finished operation.
Current setting is to flash the window indefinitely, which is a distraction, as it demands to pay attention immediately.
Other applications are usually flash only few times, which is making sense.


sdottaka:
Thank you!

Kizaemon:
Close #200

PR #236, opened 2019-12-12
Update Bulgarian translation

StoyanDimitrov:
Fix swapped translation strings. Corrected import/export outcome messages.


sdottaka:
Thank you!

PR #242, opened 2019-12-27
Consolidate FolderCmp (5)

datadiode:
- Reduce code duplication.
- Don't repeat initialization steps which are (supposedly) meant to be one-time.


sdottaka:
Thank you!

PR #243, opened 2019-12-28
Avoid some back and forth file path transcoding between UTF16 and UTF8

datadiode:
I suspect this is leftovers from ancient ANSI builds.


sdottaka:
Thank you for the PR.
However, the use of TFile () is for long file path (>256chars) support and should not be removed.

datadiode:
I see. The magic is in FileImpl::convertPath(). Quite a big journey to get the task done.

PR #244, opened 2019-12-28
GhostTextBuffer: Don't unexpectedly bring back empty lines user wants…

datadiode:
… to delete

Happens with empty lines enclosed between ghost lines.
Besides, this change seems to also save some extra messing to prevent lines from accidentally losing an EOL.


sdottaka:
This part seems to have been added by me to fix some bugs, so deleting it will recreate the bug again.
I need to remember what this bug was.

Also, it would be helpful if you could attach any files that could reproduce the problem.

sdottaka:
I have confirmed that the problem I tried to solve with the deleted line (d943136) in this PR will also be resolved in this PR.

Thank you!

PR #247, opened 2020-01-02
New branch

mesajflaviu:
I have made few modifications:
1. Added CWindowsManagerDialog class for handling open tabs with Ctrl+Tab, now the application is behave just like professional editors (Visual Studio, Notepad++, etc.) to switch and activate the open tabs.
2. m_bCloseWindowWithEsc from PropGeneral I put it **false** as default, not **true**.
3. The mainframe could be closed with Ctrl+Q (like others editors do). Also, I deactivated closing app by a simple Esc key.

The reason is that there is some modal dialog that is closing by Esc key, and if you keep Esc key a little bit longer, you close tabs and the entire app, which is not good (from my point of view). Still, the OpenView form is closing with Esc key, I cannot make it into get rid of this feature.
4. I have tried (but I didn't make it) to avoid flickering the OpenView dialog by set as TRUE the "Clipping childs" properties from this dialog, even if the flickering has disappeared, not all the controls inside of this dialog is not drawn correctly, so I give up. But I am sure this issue could be solved :)

### DIFF
--- a/Src/DirFrame.cpp
+++ b/Src/DirFrame.cpp
@@ -88,13 +88,11 @@ CDirFrame::~CDirFrame()
 {
 }
 
-
 BEGIN_MESSAGE_MAP(CDirFrame, CMergeFrameCommon)
 	//{{AFX_MSG_MAP(CDirFrame)
 	ON_WM_CREATE()
 	ON_WM_CLOSE()
 	ON_WM_SIZE()
-	ON_WM_MDIACTIVATE()
 	//}}AFX_MSG_MAP
 END_MESSAGE_MAP()
 
@@ -212,4 +210,3 @@ void CDirFrame::OnSize(UINT nType, int cx, int cy)
 	
 	m_wndFilePathBar.Resize();
 }
-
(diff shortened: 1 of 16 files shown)


sdottaka:
Thank you for the PR!

I don't want to surprise long-time users of WinMerge (I'm one of them) for steps 2 and 3, so I'll undo the fix of 2, and for 3, I'll make the behavior selectable later.

vlakoff:
Refs 2324.

vlakoff:
Refs #2324.

vlakoff:
> just like professional editors (Visual Studio, Notepad++, etc.)

Maybe Visual Studio is a ✨**professional**✨ editor, but mainly it's a bloated IDE that many people are forced to use daily (and I'm sorry for them). And that's not because it is a ✨**professional**✨ editor that we have to mimic everything they do.

Notepad++ (which I use a lot) is an editor written by a guy in his basement, I wouldn't call it a ✨**professional**✨ editor, written by a whole team following a lot of procedures.

By the way, the horrible "document switcher" can be disabled in Notepad++, and I guess it can also be disabled in the ✨**professional**✨ editor Visual Studio.

Issue #255, opened 2020-01-20
WinMerge crashes in certain circumstances, and I hope you can fix them.


265785:
When I use WinMerge, I found that WinMerge crashes in certain circumstances. This are really huge bug. I hope you can fix then as soon as possible. 
1. When I compare a lot of files (More than 1000) and the total size of them is more than 2 GB (such as lots of photos, lots of music files, lots of video files and so on). such as this: 
![20200120130725](https://user-images.githubusercontent.com/56037874/72729491-3f521d00-3bca-11ea-9441-d18c13ccb397.png)
When I compared these files in two folders and I found the different files, I used the function "copy" to make the two folders' files the same. and I copied more than 8 files. 
![0新建文本文档 - 副本 (2)](https://user-images.githubusercontent.com/56037874/72729850-24cc7380-3bcb-11ea-99c6-3a007f00b57b.png)
![0新建文本文档 - 副本](https://user-images.githubusercontent.com/56037874/72729908-4168ab80-3bcb-11ea-8ea3-415ccfcb49ae.png)
However, When I clicked the "Yes" button on the picture above, the WinMerge crashes, and it uses lots of CPU. and I don't know why. 
![0新建文本文档](https://user-images.githubusercontent.com/56037874/72730440-5b56be00-3bcc-11ea-8f30-328f67643b7a.png)
This does seriously affect normal use, especially when there are very many files.
2. When I use WinMerge to compare two files whose content is not regular. (Such as two video files, two music files, two EXE files, two ISO files and so on). 
And I previously opened them with Notepad++, and the Notepad++ didn't crashes.
![搜狗截图20200120215918](https://user-images.githubusercontent.com/56037874/72732530-db7f2280-3bd0-11ea-9a1c-a5b4ba8f2327.png)
![新建文本文档 (2) - 副本](https://user-images.githubusercontent.com/56037874/72733240-6a406f00-3bd2-11ea-9592-5b5c47aabfba.png)
It is certain that I only want to know whether the two files are same or not.
However, the WinMerge crashes for a long time when comparing these two files. especially when the two files are relatively large (More than 30 MB).  And it also uses lots of CPU.
![新建文本文档 (2)](https://user-images.githubusercontent.com/56037874/72733478-e76be400-3bd2-11ea-9704-851fa0dfb9cc.png)
This does seriously affect normal use, too. And It makes WinMerge unable to compare most types of files that is relatively large. 
[...]

PR #257, opened 2020-01-26
Prevent silent abort with File Comparison of files whose size is an exact multiple of 2^32 bytes

GreyMerlin:
Prevent silent abort with File Comparison of files whose size is an exact multiple of 2^32 bytes

* File size values that were all zero in the low 32 bits were assumed to be empty,
   and thus did not have some critical pointers initialized.  This resulted in WinMerge
   silently aborting with no message.

* WinMerge will now properly continue, but will likely display an "Out of memory" error
   message box.

* I discovered this when attempting to compare two files that are exactly 0x3700000000
   (236,223,201,280‬) bytes long.  A pointless exercise of course... Memory eventually fills up
   and an "Out of memory" error is displayed.  After clicking "OK", WinMerge terminates the
   comparison, and returns to the "Select Files or Folders" dialog, but does not actually
   release the memory until the program has completely terminated !!

* Folder Comparison with files of this extreme size does not exhibit this problem.

### DIFF
--- a/Src/Common/UniFile.cpp
+++ b/Src/Common/UniFile.cpp
@@ -241,10 +241,10 @@ bool UniMemFile::DoOpen(const String& filename, AccessMode mode)
 		return false;
 	m_lineno = -1;
 
+#ifndef _WIN64
 	unsigned sizehi = (unsigned)(m_filesize >> 32);
 	unsigned sizelo = (unsigned)(m_filesize & 0xFFFFFFFF);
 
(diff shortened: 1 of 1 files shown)


GreyMerlin:
Note ... This does **not** seem to have any relation to issue #255 (_WinMerge crashes in certain circumstances, and I hope you can fix them_)

sdottaka:
Thank you!

PR #269, opened 2020-02-23
Include Excel Addin files (*.xlam)

cswartzvi:
The current CompareMSExcelFiles plugin does not identify Excel addin files (*.xlam) as an allowed extension. While the user can manually specify the unpacker plugin when `Manual Unpacking` is enabled, this is not possible with `Automatic Unpacking`. This pull request adds the *.xlam extension to the CompareMSExcelFiles and removes *.xlax (which I am assuming was a typo).


sdottaka:
Thank you!

PR #271, opened 2020-02-25
Update Russian.po

wvxwxvw:
Translated the missing text, corrected some things, and removed the extra lines in accordance with POT.


sdottaka:
Thank you!

PR #292, opened 2020-03-21
Galician and Spanish translations update

qosobrin:
I have just updated the Galician and Spanish translations for the InnoSetup and the Shell Extension files.


sdottaka:
Thank you!

PR #293, opened 2020-03-22
Typographical and encoding corrections to Galician and Spanish Shell Extension translations

qosobrin:
I made a couple of corrections to the Galician and Spanish Shell Extension translations. Everything should be OK now. Sorry for the noise.


sdottaka:
Thank you!

PR #312, opened 2020-04-15
"Patience" diff refers to the card game

tkoester:
For a description of the algorithm see https://blog.jcoglan.com/2017/09/19/the-patience-diff-algorithm/
The lines are to be split and sorted just like the card decks in solitaire, hence the name.


sdottaka:
Thank you!

PR #313, opened 2020-04-17
Galician WinMerge translation complete

qosobrin:
I have just finished the translation into Galician of the WinMerge interface. I have done a quick quality check to correct all the issues that I could find, but I am sure that some more issues will come up in the future. I have checked that the underlined letters are unique for every menu but, due to the complexity of some menus that change options according the context, some more tests are needed.

Some problems have arisen during the translation that I think should be treated as bugs but I will open a bug issue to treat them later. 

Best regards.


sdottaka:
Thank you!

Issue #283, opened 2020-03-09
Create a patch file from directories


jonadem:
When the "Generate patch" is pressed (under "Tools") with a directories diff, WinMerge returns a dialog "Cannot create a patch file from directories". Could you implement this ?

sdottaka:
I don't think it will be implemented unless someone makes a PullRequest.

Currently, WinMerge does not allow you to create a patch by specifying a folder. However, you can select multiple files to create a patch.

If there are subfolders in the folder, it may help to enable the 'Include Subfolders' option or disable tree mode.

sdottaka:
implemented by PR #331

PR #331, opened 2020-05-03
Add a feature generating a patch from directories (#283)

tjmprm77:
Add a feature generating a patch from directories in relation to Issue #283.
- Add a feature generating a patch from directories.
- Remove some processing that abort patch generation when attempting to generate a patch from directories.
- Change to open a file or directory selection dialog when clicking [Browse...] buttons in [Patch Generator] dialog.
- Change the behavior if the input is a binary file as follows:
  - Processing is not aborted but continued.
  - "Cannot create a patch file from binary files." message is shown once only at the first time per patch generation process.
  - Even if "Cannot create a patch file from binary files." message is shown, if patch generation succeed for other regular files, "Patch file successfully written." messages is shown.


sdottaka:
Thank you!

jimsmithm3:
@sdottaka Can you think of a reason when I create the patch file output that the filenames are not showing up?   

I am doing what you showed in #1753 but still don't see the filenames if I do it for a single file or a folder and outputting to a file or the clipboard.

Using 2.16.30.0  

[pat1560.txt](https://github.com/WinMerge/winmerge/files/11492846/pat1560.txt)

Thanks for the great tool.

sdottaka:
@jimsmithm3
If you select something other than Normal in the Style combobox of the Patch Generator dialog, the file name will be output.

image

jimsmithm3:
That indeed was the easy fix and I see them now.  Thanks so much for the quick response.

PR #335, opened 2020-05-08
2.16.6 Turkish translation update

kayazeren:


sdottaka:
Thank you!
The Translations part is missing, so I'll put it back later.

PR #337, opened 2020-05-08
2.16.6 Turkish translation

kayazeren:
Small fixes after full review


sdottaka:
Thank you.

By the way, it seems that you are not only fixing the Turkish language strings, but also the English strings as well. 
In this case, when WinMerge is built, the following script will automatically delete the Turkish string corresponding to the changed English string, wasting your work.

https://github.com/WinMerge/winmerge/blob/master/Translations/WinMerge/UpdatePoFilesFromPotFile.ps1.


This time, I'll deal with it. Next time, if you want to modify the English string, please modify the following file.

https://github.com/WinMerge/winmerge/blob/master/Src/Merge.rc

kayazeren:
Thanks for notice

Some translations did not appear on WinMerge and I thought that original
strings on my translation file were not correct
It seems like I took the wrong way between 2.16.4 and 2.16.6. So I will
cross-check by pot file and WinMerge 2.16.6 again.


Takashi Sawanaka <notifications@github.com>, 8 May 2020 Cum, 13:52
tarihinde şunu yazdı:

PR #366, opened 2020-06-27
Change to disable some menu items that should not be executed for directories, when directories are selected in the folder compare window.

tjmprm77:
Change to disable some menu items listed below, when directories are selected in the folder compare window. 
- Context menu
  - "Compare As" > "XML"
  - "Compare As" > "Binary"
  - "Compare As" > "Image"
- Menu Bar
  - "Plugins" > "Edit with Unpacker..."

Because the current version of WinMerge does not give appropriate behaviour when selecting above menu items with selecting directories in the folder compare window, and because they do not seem to be expected to execute for directories.


sdottaka:
Thank you!

PR #376, opened 2020-07-17
Update DirViewColItems.cpp

yfdyh000:
I hope this fixes the problem that it is not localizable. I also don't know why it has an extra point.
https://github.com/WinMerge/winmerge/blame/363854a8de77e0ff78822782306783baea5ecbaf/Translations/WinMerge/English.pot#L2395


sdottaka:
Thank you!

PR #381, opened 2020-07-24
Fix the problem that not 3 panes but 2 panes file compare window displays when executing "Compare As" > "Binary" or "Compare As" > "Image" in the context menu in the folder compare window with selecting 3 items.

tjmprm77:
Fix the problem that not 3 panes but 2 panes file compare window displays when executing "Compare As" > "Binary" or "Compare As" > "Image" in the context menu in the folder compare window with selecting 3 items.


sdottaka:
Thank you!

PR #386, opened 2020-07-28
Turkish translation update 2.16.7

kayazeren:
Turkish translation update 2.16.7


sdottaka:
Thank you!
It seems that "https" has changed back to "http", and the information of the translator has disappeared, so I will correct it later.

Issue #316, opened 2020-04-21
GUI glitches/bugs


vertigo220:
Some minor issues with the GUI, low-priority:

1) Options > Compare > General > Diff algorith drop-down box is incomplete until hovered over. When first entering the screen, the bottom outline of the box doesn't show, but when moving the mouse over it, it becomes visible.
2) About WinMerge window contents disappear when Alt-Tabbing. I opened it to get the ver # and switched here to enter it, and when I switched back the about window was empty. It happens every time.
3) WinMerge crashed when clicking the Options > Compare > Binary > View settings... button. Relaunching it and clicking the button again worked, so it's not consistent.

v2.16.6.0, Win10x64

sdottaka:
Thank you for the report.

I've been fixed 1. and 3.  by commit 35f4d061838c35c2a0b59a85e49f2ecd97b27809 and da43ee96b59cbd4d3dedd7e4b3bfacca449faec9

However, I couldn't reproduce 2.
Please let me know if there is a way to reproduce it.

![316](https://user-images.githubusercontent.com/98126/79922093-44936200-846e-11ea-9021-0359eb7e8cd8.gif)

vertigo220:
All I do is launch WinMerge, open the about screen, and press Alt and all the text on the screen disappears. Trying it out again right now to see if I could figure out more about it led me to discover I don't even have to Alt+Tab, just pressing Alt does it. It's possible it's because I uninstalled an older version and didn't restart even though it said I needed to, then installed the latest version, but it seems strange that would have such an effect. I can see if it's fixed after I restart, but I'm not sure when that will be.

On a side note, I created an account and tried to make several posts on the forum last night, and nothing is going through as they're awaiting moderation. If you happen to be an admin there or know one, I'd appreciate if that could be done.

sdottaka:
> All I do is launch WinMerge, open the about screen, and press Alt and all the text on the screen disappears. Trying it out again right now to see if I could figure out more about it led me to discover I don't even have to Alt+Tab, just pressing Alt does it. It's possible it's because I uninstalled an older version and didn't restart even though it said I needed to, then installed the latest version, but it seems strange that would have such an effect. I can see if it's fixed after I restart, but I'm not sure when that will be.

Thank you for the info.
I am now able to reproduce the bug.
I couldn't reproduce the bug because I was showing the About dialog with a keyboard shortcut.

I've fixed the bug with commit ed69d2c.

[...]

PR #392, opened 2020-08-05
Rework the fix for Github issue #316: GUI glitches/bugs #2

datadiode:
(Not really relevant, but gets rid of some minor occasional flicker, and might be technically preferable from a pedantic perspective.)


sdottaka:
Thank you!

PR #394, opened 2020-08-07
Replace outdated list of developers in AboutBox with ascii-art gnu fr…

datadiode:
…om FSF.

The list as is no longer tells a valid story.
Updating the list (as happened occasionally in the past) imposes a risk of snubbing someone.
Therefore, I propose to remove it in favor of some piece of ASCII art.


gerundt:
At least Takashi is still active! And maybe "Jochen Neubeck"? ;)

Its ok for me to lose my name in the about dialog.

sdottaka:
I have no objection. Thank you for PR.

PR #397, opened 2020-08-09
WildcardDropList: Avoid the String instance as it could throw std::ba…

datadiode:
…d_alloc


sdottaka:
The reason I introduced patternsInList was to remove any duplication in the user-added patterns.
However, I accept this PR as it is not very important to remove the duplication.

PR #400, opened 2020-08-09
Remove duplicate filename patterns without relying on WildcardDropList

datadiode:
This works also when user edits patterns without opening the drop list.


sdottaka:
Thank you!

PR #421, opened 2020-08-30
Fixed the problem that the input to File1 or File2 comboboxes of the Patch Generator dialog is not applied when the Patch Generator dialog is opened by selecting multiple files.

tjmprm77:
In Current version, the input to File1 or File2 comboboxes of the Patch Generator dialog is not applied when the Patch Generator dialog is opened by selecting multiple files.

How to reproduce:
1. Select multiple files in the folder compare window and run "Generate Patch..." from the menu.
2. In the Patch Generator dialog, enter the files again in the File1 and File2 comboboxes and click the "OK" button.
3. The patch of the file selected in step 1 is generated instead of the file specified in step 2.

The cause is that the contents of the File1 and File2 comboboxes are read only if single files registered when the "OK" button is pressed in the Patch Generator dialog.

This PR fixes the problem by clearing the internal item list of the CPatchDlg class when multiple items are registered in the internal item list, and the File1 or File2 comboboxes of the Patch Generator dialog is changed.


sdottaka:
Thank you!

PR #424, opened 2020-09-05
Tweak output from BuildInstaller.cmd and BuildArc.cmd

gerundt:
This PR adds a *Build WinMerge%PLATFORM%.iss with Inno Setup 5* echo to `BuildInstaller.cmd` and hides two *1 file(s) copied.* from `BuildArc.cmd`


sdottaka:
Thank you!

PR #435, opened 2020-09-22
Fix the problem that the comparison result is displayed one line off in the file compare window if a sync point is added with the first line selected.

tjmprm77:
The comparison result is displayed one line off in the file compare window with Release build if a sync point is added with the first line selected on one side and the non-first line selected on the other side.

And when the same operation is performed with Debug build, an assert error occurs at the code "`assert(curDiff.begin[0]+extras[0] == curDiff.begin[1]+extras[1]);`" in the function "`void DiffList::GetExtraLinesCounts(int nFiles, int extras[3])`".

Cause:
WinMerge separates the files to be compared at the sync point, makes a diff for each separated file by diffutils, and then merges it.
If a sync point is added on the first line, the first separated file is an empty file.
If an empty file is specified as the input file, diffutils returns the result that a one-line diff exists.
Due to the one-line diff, the comparison result is displayed one line off in the file compare window.

This PR fixes the problem by Correcting the comparison results made by diffutils if the first file separated by the sync point is an empty file.

P.S. Thank you for adding me to Contributors.txt.


sdottaka:
Thank you!

PR #457, opened 2020-10-17
Fix the problem accessing the text buffer out of range if a sync point is added by selecting a ghost line that is after the last block.

tjmprm77:
The current version of WinMerge accesses the text buffer out of range if a sync point is added by selecting a ghost line that is after the last block.
This PR fixes this problem by doing the following:
- Show an error message and cancel the process adding a sync point if a sync point is added by selecting a ghost line that is after the last block.
- Add some assertions.


sdottaka:
Thank you!

I feel that the error message "Cannot add a synchronization point after the last block" is a bit inappropriate, so I may fix it later.

tjmprm77:
Thank you for fixing the error message.
I also think the fixed error message is more appropriate.

Issue #279, opened 2020-03-05
"copy selected text only" - bug or feature?


SamuelPlentz:
Thanks a lot for your work!

### First issue:
In previous versions (2.16.2.0 or 2.16.0.0) when I selected a line by double clicking it and then used the shortcut "alt + left arrow" **the whole line** was copied to the left.

In the current version (2.16.6.0) when I select a line by double clicking it and then use the shortcut "alt + left arrow" **only the selected text** is copied to the left.

Is it a bug or a feature? If it is a feature I really would like to have an option to copy full lines only. I know about triple click, but my workflow doesn't get used to it ;)

### Second issue:

In the following case "b2" is copied to the left in front of the line "c1". I would have expected it to copy to a separate line. (it works as expected when I select the full row)

![merge](https://user-images.githubusercontent.com/1890854/75957871-4b184b00-5ebb-11ea-8cfa-1de0629cef28.png)

SamuelPlentz:
Should be this part of the code:
https://github.com/WinMerge/winmerge/blob/98f1c722244667c426a864a179998e3b1f751359/Src/MergeEditView.cpp#L1841-L1889

SamuelPlentz:
So issue 1 would be to have an option to also use this part:

```cpp
CWaitCursor waitstatus; 
pDoc->ListCopy(srcPane, dstPane, currentDiff); 
```

instead of this part:

```cpp
int firstDiff, lastDiff, firstWordDiff, lastWordDiff; 
GetFullySelectedDiffs(firstDiff, lastDiff, firstWordDiff, lastWordDiff); 
if (firstDiff != -1 && lastDiff != -1) 
{ 
    CWaitCursor waitstatus; 
    pDoc->CopyMultipleList(srcPane, dstPane, firstDiff, lastDiff, firstWordDiff, lastWordDiff); 
} 
```

Column selection is not affected and could stay as it is.

sdottaka:
> First issue:
> Is it a bug or a feature

It is a feature. 
Are you double-clicking to select a difference?
In that case, I recommend using Alt + Enter key.
However, you can use Alt + Left / Right or Right key to merge the differences at the cursor position without selecting a difference.

>Second issue:
I could not reproduce the bug as follows:
![secondissue](https://user-images.githubusercontent.com/98126/76034659-e45e6480-5f82-11ea-84e0-7b8b72bffa1d.gif)

Can you tell me if you find the conditions to reproduce?

[...]

PR #459, opened 2020-10-17
always copy "full line" instead of "selected text only" - option CopyFullLine

SamuelPlentz:
Like described in #279:

Restore old copy behaviour (always copy "full line" instead of "selected text only"), with a hidden option.
Name of the option is: `CopyFullLine`
Default value is: `false` (deactivated)

For now, I didn't add something in the GUI, in #279 @sdottaka wrote:

> If few people need this option, I'd like to make it a hidden option, but if it's more, you'd better add a checkbox to the Options dialog.I don't know whether many people need this option or not. I'll leave it to you.

If it should be added to the GUI it should be below "Preserve original EOL chars" on the Editor page. But the Editor page lacks room for it.

If there will be an option it would be good to name it: "Always copy full lines instead of selected differences only"

Should I document the hidden option somewhere? Is there a documentation for hidden options?

### DIFF
--- a/Src/CompareOptions.cpp
+++ b/Src/CompareOptions.cpp
@@ -21,6 +21,7 @@ CompareOptions::CompareOptions()
 , m_bIgnoreEOLDifference(false)
 , m_diffAlgorithm(DIFF_ALGORITHM_DEFAULT)
 , m_bIndentHeuristic(true)
+, m_bCopyFullLine(false)
 {
 }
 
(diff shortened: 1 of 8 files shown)


sdottaka:
Thank you!

I don't know when it can be implemented, but I would like to add the ability to copy only the current line or only the selected lines to the left or right.
Along with that, I would like to be able to select the behavior of ALT + left / right keys when text is selected as shown below.

-Copy the selected inline differences to the left or right (current behavior)
-Copy the selected diff blocks to the left or right (2.16.2 or earlier)
-Copy the selected lines to the left or right (future)

As a result, there may be option renaming and documentation changes, and modifying the documentation now may be useless.

SamuelPlentz:
Thanks for accepting this. An option like this sounds like a good idea.

sdottaka:
PR#2224 added an option to select the behavior when copying to another pane with "Copy to right/left" while text is selected.

PR #476, opened 2020-11-14
Fix the problem that "Goto...", "Goto Line <line>" and moving by left click on the location pane do not work appropriately when the Diff pane is active in the file compare window.

tjmprm77:
Fix the problem that "Goto...", "Goto Line <line>" and moving by left click on the location pane do not work appropriately when the Diff pane is active in the file compare window.

Activates the File pane and go to the specified line if the above operation is performed when the Diff pane is active.


sdottaka:
Thank you!

PR #497, opened 2020-11-30
Update manual

Mr-Update:
- Spelling corrected
- "Getting and installing WinMerge" completed
- "Click Next in the Welcome screen." deleted
- "Create a Quick Launch shortcut" deleted
- Order of plugins adopted
- Shortcuts added

- I found the sentence "At any time, you can click **Defaults** to reload the installed color scheme." in "Options and configuration" -> "Colors page", "Syntax Colors page" and "Folder Compare Colors page".
But in the 3 sections of the options dialog there is no **Defaults** button.
Is there a reason for this?


sdottaka:
Thank you!

This is because the color scheme is now selectable.
When the default button was pressed, only the colors of that page had to be reverted to the selected color scheme, but I removed that default button for ease of implementation.
Instead, select the Color Scheme combo box again to return to the default color.

PR #523, opened 2020-12-26
Fixed the problem that the language setting is not applied to the context menu of the MDI tab when the language setting is changed.

tjmprm77:
In the current version of WinMerge, the language setting is not applied to the context menu of the MDI tab when the language setting is changed.
It is because the context menu strings of the MDI tab are set based on the language setting at that time only at the first display for each tab, and the set strings are used from the second time.
This PR Fixes to set the menu strings based on the language setting every time the context menu is displayed.


sdottaka:
Thank you!

PR #538, opened 2021-01-04
Galician translation update

qosobrin:
I have just updated the Galician translation. I would appreciate if you could remove former translator (Medulio) and add me to the list of translators since I did this translation _ex novo_ The reason for this is because the translation he provided had so many errors that it was easier for me to start again than to correct.

Thank you very much.

Best regards.


sdottaka:
Thank you!

sdottaka:
I added your GitHub username on line 6 of Galician.po. when puting names here, it will be reflected on the Translation Stauts page. (http://htmlpreview.github.io/?https://github.com/WinMerge/winmerge/blob/master/Translations/TranslationsStatus.html)
If you want to change it, please modify Galician.po and make a pull request.

(I want to keep the name of the person who contributed in the past, so I keep the name of Medulio)

PR #544, opened 2021-01-08
A new feature "Ignored Substitutions"

EugeneLaptev:
This is a feature to facilitate a programmer workflow. If you deal with a big code base and happen to rename a symbol you may get hundreds of individual changes that are all the same. To let you focus on meaningful changes and filter out those symbol renames I've added another sort of filter called "Ignored Substitutions".

You can edit this filter in a standard way or you can right click on a change and select "Add to Ignored Substitutions", after which those changes will be ignored by the vie panes.

(Sorry is something is trivially wrong- this is my very first experience with github.)


Mr-Update:
This is a very good idea. 😊

I can use this feature very well for programmes with German language files that cannot be improved by a pull request or an e-mail.

Then I maintain my own language file and can hide a "funny" or suddenly wrong translation of a word when comparing with WinMerge.

sdottaka:
Thankl you!

sdottaka:
It's a good feature, but it doesn't seem to work as expected as below.

image

I also want to be able to replace the PrediffLineFilter.sct plugin, which already has similar functionality, so I'd like to make some changes.

However, I want to release a new version this month, so this feature itself may not be included in the new version.

EugeneLaptev:
Yes, thank you. I'll fix that a bit later (Monday). The problem is probbaly
because it's the last line. I could not reliably figure out how to detect
that.

Good luck
Eugene

sdottaka:
@EugeneLaptev

The Ignored Substitutions dialog has been modified as follows:

image

By the way, why did you name it Ignored Substitutions instead of  "Substitution filters"?

I feel that the name is not consistent with other File filters and Line filters.

EugeneLaptev:
Lovely!

"Substituion filter" is a good idea, it was on my short list. I also
considered "Rename filters", as this was what the feature was
specifically meant to be used for, but was not punchy and clear
enough.

In my coding practice I am a naming nazi - it's very important to name
things well. But the verb "filter", likely because of my engineering
background, has ambiguous meaning. The verb sometimes means "filter
out" (="throw away") or the opposite "allow in" (="pass through"),
with roughly the same frequency. So the verb alone does not
immediately tell you what's going on. That's why I went for explicit
term "ignored", which conveys the process under the hood. However, for
the big things such as WinMerge it's important to have consistency
throughout. So please rename to "Substitution filters" if you like it
more.

Good luck
Eugene

sdottaka:
Thank you for your answer.
If only this feature exists, the name "Ignored Substitutions" might be appropriate, but I would like to change it to "Substitution Filters" as it is inconsistent with other terms.

PR #545, opened 2021-01-08
Various fixes to testing

GreyMerlin:
## ?? Bring FreeImage submodule into sync ??
* I have no idea why **Git** thinks this is necessary.  I'll have no complaint if this commit is rejected.

## Fix syntax errors with **`Test`** solution configuration
* Recent additions have been made to the parameters for the `GetMainFrame()->FileNew()` procedure, but not incorporated into the code for the **`Test`** configuration.

## Fix Failures with **GuiTests** `InstallerTest` module
* If the `...-x64-PerUser-Setup.exe` file is missing from the `Build/` folder tree, then **GuiTests** fails with
  a)  A message box (from deep within the `system()` function) suggesting a file name misspelling, plus
  b)  Seven `ASSERT` failures that have to be individually, manually Ignored.

* This patch allows **GuiTests** to detect the missing file itself and cause a proper **`GTEST_FAIL()`**
action.

* The filename for the required `...Setup.exe` file is upgraded from version `2.16.7.0` to `2.16.8` (the latest
publicly available setup file that I could find).

### DIFF
--- a/Src/Test.cpp
+++ b/Src/Test.cpp
@@ -470,10 +470,10 @@ TEST(ImageCompareTest, Open)
 TEST(FileMenu, New)
 {
 	CFrameWnd *pFrame;
-	GetMainFrame()->FileNew(2);
+	GetMainFrame()->FileNew(2, CMainFrame::FRAMETYPE::FRAME_FILE, false);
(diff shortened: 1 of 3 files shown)


sdottaka:
Thank you!

PR #546, opened 2021-01-08
Update Manual

Mr-Update:
- Spelling corrected
- Middle elements added
- Filter folder in Options and configuration updated
- Ins/Del Detection in Compare Images added

How does "Ins/Del Detection" work?


sdottaka:
Thank you!

"Ins/Del Detection" applies the Diff algorithm to detect the insertion and deletion of images by treating the lines of the image as if they were lines of a text file.

ins_del_dectection

PR #549, opened 2021-01-11
Fixed mistakes in Ignored Substitutions

EugeneLaptev:
I've fixed the two issues that you've identified, which are seen on the picture:
https://user-images.githubusercontent.com/98126/104092948-9dd84580-52ca-11eb-80f8-fa03742f2a1c.png

(Also please check that for that particular case on the picture the highlighting of the line difference on the character level works correctly.  To me it looks like the difference is lit on the world level: "IDENTLEVEL" - even if you ask to highlight on the character level.)

I've also created two commits with similar descriptions, which I'd fuse into one commit (with the last message), but my git skills are not sufficient for such operation.


sdottaka:
Thank you!

I may make the following changes later.

- Add the "Match cases" checkbox
- Match terms to the terms in the replace dialog.
   e.g.
   On one panel-> Find what
   On the other panel-> Replace with
- Move the "Completely unhighlight the ignore change" checkbox to the options dialog.
    (This function should also work with line filters etc.)
- Allow regular expressions \1 and \2 to be specified on the "On the other panel" side.
   e.g.
   On one panel: func\(([^,]+), ([^)]+)\)
   On the other panel: func(\2, \1)

    So I might remove the "Ignore changes in both directions" checkbox.

EugeneLaptev:
These are excellent ideas! I did not need case sensitivity as I mainly use
WinMerge for code, but other people would likely want to be able to ignore
case.
Thanks guys for providing an excellent tool for free. Quite a few of my
colleagues use it.

PR #556, opened 2021-01-13
Some tweaks for translations status

gerundt:
`GetTranslationsStatus.py` creates now also a file in Markdown. Maybe this would be the best file for the Git repository?

MD und HTML reports has now also a column with a percentage value.


sdottaka:
Thank you!

Issue #475, opened 2020-11-13
Feature Request: Move To Next File option while comparing files


jschof:
While comparing files when you reach the last difference and try to go to the next difference you are prompted with the option to move to the next file. This is fantastic but I would like to go to the next file at any time.

I don't see a menu item for this feature. It would be nice if it could be added there and optimally a button added as well.

To complete the feature it would be nice to also be able to navigate to the 1st, previous, last file similar to what you can do with differences.

Thanks!

sdottaka:
If you don't want to see the message box, please check the [Don't ask this question again] checkbox above the message box and click [Yes] button.

![image](https://user-images.githubusercontent.com/98126/99136311-859bcf80-2667-11eb-9737-9db4811aa364.png)

jschof:
Thanks but I have no problem with the prompt. I just want to be able to skip to the next file without having to go to the bottom of the file first and then going to the next difference. 

I understand that I can get the result i want with the click of 2 buttons but it would be much friendlier to just have to click on a single button to achieve this.

sdottaka:
sorry. I misread it.

jschof:
Np, I appreciate you trying to help nonetheless.

sdottaka:
Implemented by PR #561

PR #561, opened 2021-01-17
Feature Request: Move To Next File option while comparing files #475 …

ranjwarrior:
…- Navigating between files in file comparison mode. Will attempt to move to first/next/prev/last files only, ignoring/skipping directories.
(Note: Keyboard navigation, or any other missing options can be added if required)


sdottaka:
Thank you!

ranjwarrior:
Thanks for merging. Happy to contribute.

jschof:
Thanks @ranjwarrior it's appreciated!

mrnams:
How to disable it? I want to disable auto move to next file

sdottaka:
By default, it does not move to the next file automatically and displays a message box asking "Do you want to move to the next file?".
You probably checked the "Don't ask this question again" checkbox in the message box to prevent it from appearing.
If you want this message box to appear again, uncheck the checkbox in the [Message Boxes] category in the Options window.
Or, I think the Answer column in this [Message Boxes] category is set to "Yes", so if you set it to "No" and leave the checkbox checked, it will not move to the next file.

image

PR #570, opened 2021-01-22
Galician translation update

qosobrin:


sdottaka:
Thank you!

sdottaka:
I changed the term "Ignore Substitutions" to "Substitution Filters" and lost some translations. I'm sorry that your translation was lost.

PR #593, opened 2021-02-01
Make all OK strings same case

StoyanDimitrov:
Some translation tools are not happy about existence of both "&OK" and "&Ok". Besides that it let's reuse!


sdottaka:
Thank you!

PR #596, opened 2021-02-01
Fix memory leaks

thehlopster:
Allocated array should be deallocated with `delete []` Deallocating it with `delete` can cause memory leaks.


sdottaka:
Thank you!

Issue #612, opened 2021-02-08
Expression is always true


thehlopster:
https://github.com/WinMerge/winmerge/blob/a9b91106cbd04ddf48a0c895e1ab3de34ce9654d/Src/PatchHTML.cpp#L434-L436
since `spaces` is `unsigned` it is never lower than 0, so this check is unnecessary and can be removed

PR #613, opened 2021-02-08
Expression is always true

thehlopster:
#612 since spaces is unsigned it is never lower than 0, so this check is unnecessary and can be removed


sdottaka:
Thank you!

PR #624, opened 2021-02-11
Fix an issue where the syntax highlighting scheme is not applied to the split second File pane and the Diff pane when changing it.

tjmprm77:
Register the syntax highlighting scheme in the second split File pane and the Diff pane, when it is changed.


sdottaka:
Thank you!

Issue #622, opened 2021-02-11
Single instance mode does not work when launched from Visual Studio 2019


CrendKing:
I have a git repo. The relevant git config line is
```
[difftool "winmerge"]
	cmd = WinMergeU.exe -s $LOCAL $REMOTE
```
WinMerge registry is completely empty and all settings default.

I open this repo in Visual Studio 2019. The "Git Changes" tab shows all the changed files. When I diff the first pair of files, everything works fine. When I diff the next pair, WinMerge shows error saying "Cannot open file C:\Users\CrendKing\AppData\Local\Temp\TFSTemp\vctemp12345_67890.test.xxxxx.cpp". The exact same setup works with Beyond Compare. Also works if I don't use `-s` or single instance mode in options, but then it will open two windows instead of adding a new tab, which I prefer.

When I go to the temp directory, I can only see the first temp file. It seems like VS2019 creates that temp file, send to WinMerge, but WinMerge terminates that new process immediately without any wait, VS2019 thought the diff is done thus clean up the temp file, but then the old WinMerge process pick up the diff request, unable to find the temp file.

In contrast, Beyond Compare has two executables: BCompare.exe is the GUI and actual diff tool and BComp.exe is a loader. Git by default prefers BComp and it has ability to send request to existing BCompare process. Thus we could have several BComp process, each waiting for diff to finish, but only one BCompare process.

So can WinMerge somehow make single instance mode work with tools like Visual Studio 2019?

sdottaka:
I have added a new command line option /sw. This option limits the WinMerge window that runs like /s to one, but waits for WinMerge that is already running to terminate.

PR #637, opened 2021-02-18
Update French.po

slolo2000:
Translate sentence added with "Fix GitHub issue #622": Single instance mode does not work when launch…


sdottaka:
Thank you!

Issue #463, opened 2020-10-20
Feature request: Add "Copy current [line/individual change] [to/from] [Left/Right]" in Merge functions


dmz73vu:
Currently Merge functions exist to Copy [from/to] [left/right] but those functions operate on the whole block of changes.
Often there are individual lines or changes that should be copied and the rest of the changed block should be ignored.
It would be nice to be able to copy those changes without manually editing the file.

sdottaka:
Implemented by PR #642

PR #642, opened 2021-02-21
Add "Copy Selected Line(s) to/from Left/Right" feature ( #463)

elsonwei:


sdottaka:
Thank you!

jmonachesi:
When double-clicking to highlight a block to transfer from Left to Right (Alt-Right) or viceversa, usually some portion of text is selected. This makes a conflict with the most used feature of WinMerge which is to transfer blocks of differences. 

It would be better if this new feature could be deactivated in the options (and off by default) because as it is a niche utility and I believe that long time users of WinMerge will not appreciate changing the default behaviour. 

Thank you.

PR #719, opened 2021-04-06
Add C# 6.0  reserved words to style list

digocesar:
`when` was include in C# 6.0 for catch and in C# 7.0 for switch/case.
https://docs.microsoft.com/pt-br/dotnet/csharp/language-reference/keywords/when


sdottaka:
Thank you!

PR #724, opened 2021-04-10
Update PHP syntax highlighting keyword list.

tjmprm77:
- Added some keywords
- Modify the list grouping as described in the PHP documentation. (https://www.php.net/manual/en/reserved.php)


sdottaka:
Thank you!

PR #728, opened 2021-04-14
Update Brazilian.po

Felipefpl:
Updated the translation here because the .po/.pot files found here are reported as corrupted by poedit https://winmerge.org/translations/?lang=en


sdottaka:
Thank you!

PR #729, opened 2021-04-14
Update Brazilian.isl

Felipefpl:
Updated the installer here since comparing the files from the site wont do since it's a mess.


sdottaka:
Thank you!

Issue #248, opened 2020-01-04
One suggestion: You should save the settings in an "ini" file instead of in the registry, especially the zip version.. 


265785:
The portable versions of WinMerge is not portable anymore.
When I download the portable versions (Zip version) of WinMerge on your website, and I place it on my portable drive, I found that WinMerge saved the settings and so on in the registry. And the settings is unabie to delete if I don't browse the registry by myself. 
![搜狗截图20200104162155](https://user-images.githubusercontent.com/56037874/71762747-60e6af80-2f0e-11ea-8ad0-e291029c77e5.png)
![搜狗截图20191203010241](https://user-images.githubusercontent.com/56037874/71762702-bff7f480-2f0d-11ea-9036-78517ce07d98.png)
So the settings in the registry becomes rubbish when I delete the portable version of WinMerge. and it is hard to delete, Only if I browse the registry or reinstall the system. 
Also,I found that WinMerge is impossible to place in a portable drive. If I place the WinMerge in a portable drive, the settings is unable to save in the portable drive. If I change the computer, the settings will be lost, and it produces rubbish in the previous computer's registry. I have to make setting backups and I have to import the settings each time I change one computer. that's too much trouble. 
In order to solve this problem, I can give you one suggesstion: 
You should save the settings in an "ini" file instead of in the registry. and the "ini" file can be placed in the root folder of your program. Also your software should directly read the settings in the "ini" file directly instead of reading the registy. 
Like this (XnView): 
![搜狗截图20200104160500](https://user-images.githubusercontent.com/56037874/71762712-daca6900-2f0d-11ea-9690-6ea3aea07696.png)
And this (Everything): 
![搜狗截图20200104160749](https://user-images.githubusercontent.com/56037874/71762716-fa619180-2f0d-11ea-8d31-ff8c2baa7177.png)
This can not only make your software portable, but also make the settings backuping much more easily. 
I hope you can receive my suggestions. Thank you!

MailYouLater:
Note: Saving settings in the registry is fine for a normal install, but there are people who prefer portability, so it would be very nice if there was an option to save the settings to an ini instead.
WinMerge Portable has a workaround for this issue in that it has a launcher which loads the settings into the registry, launches WinMerge, waits for WinMerge to exit, then moves the settings back to disk, but it would be much nicer if it didn't have to have a launcher which waits around to clean up after it.

[...]

PR #750, opened 2021-04-24
Possibility to store settings in INI file (#248)

grzesie:
Hi.

Here is first code that allows to store setting in INI file. According to issue #248 it checks if INI file exist - if it exist, uses it. Otherwise use previous register mechanism.

### DIFF
--- a/Src/Common/CIniOptionsMgr.cpp
+++ b/Src/Common/CIniOptionsMgr.cpp
@@ -1,12 +1,12 @@
 /**
- * @file IniOptionsMgr.cpp
+ * @file CIniOptionsMgr.cpp
  *
  * @brief Implementation of Ini file Options management class.
  *
  */
 
 #include "pch.h"
-#include "IniOptionsMgr.h"
+#include "CIniOptionsMgr.h"
 #include "OptionsMgr.h"
 #include <Windows.h>
 #include <codecvt>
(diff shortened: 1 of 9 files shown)


sdottaka:
Thank you!

I will release version 2.16.12 this month, so I will include this Pull Request after the release.

After applying this Pull Request, I may change the following parts later.

- If using std::filesystem causes an unacceptable increase in binary size or does not work on Windows XP, I will change it to use the Poco library or Win32 API instead.

- Since the behavior has changed due to comment out (ToDo: part), I want to restore it.

- I want to rename the class of IniOptionsMgr to CIniOptionsMgr to match CRegOptionsMgr.

- Since there is a concern that performance will deteriorate using INI file instead of the registry, I would like to introduce an asynchronous writing mechanism.

grzesie:
> * Since the behavior has changed due to comment out (ToDo: part), I want to restore it.

Maybe it should be changed to this:
`if (typeid(*pOptions) == typeid(CRegOptionsMgr))
	{
		static_cast<CRegOptionsMgr *>(GetOptionsMgr())->CloseKeys();
	}`

Should i change this and class names?

sdottaka:
If you have time, I would appreciate it if you could make that changes.

grzesie:
Name changed, also removed this TODO and wrapped it in IF. However 1 comment here:
This is in OnIdle method - it is executing many times. I don't think closing keys should be executed so many times during program execution. But maybe i am not aware of something and it is must have...

sdottaka:
Thank you for the update.

There is an environment where it takes a long time to open the registry key, so keeping the registry key open to speed up continuous access to the registry.
However, I don't want to keep the registry key open all the time, so closing the registry key when WinMerge is in the Idle state.
CloseKeys() is called many times, but if the registry key is already closed, nothing will happen, so I think it's okay. Of course, Pull Requests are welcome if there is another good way.

PR #751, opened 2021-04-24
Fix PHP syntax highlighting.

tjmprm77:
Fix an issue where the string after the block whose color index is COLORINDEX_NORMALTEXT or COLORINDEX_USER1 is not highlighted correctly.
(The processing of the fixed part is now the same as CrystalLineParser::ParseLineCJava() with this fix.)

Current version：
current

Fixed version:
fixed


sdottaka:
Thank you!

PR #753, opened 2021-04-26
autoit.cpp - #CS #CE  support

mlipok:
Added #CS #CE support .... as so far only lower case #cs #ce was supported.


sdottaka:
Thank you!

This commit undoes the following indentation changes, so I'll change them again.
If you have any complaints about this, please let me know.

https://github.com/WinMerge/winmerge/commit/46546b0aceb516696662be481218c207e0725527

mlipok:
If have no any complaints about this, just my oversight.

PR #763, opened 2021-04-30
Fix an issue where a message is not translated.

tjmprm77:
Fix an issue where the message "The number of automatically merged changes: %1\nThe number of unresolved conflicts: %2" is not translated.


sdottaka:
Thank you!

PR #782, opened 2021-05-07
Fix PHP syntax highlighting.

tjmprm77:
Fix the following PHP syntax highlighting issues:
- The second and subsequent lines of the string enclosed in single quotes are not displayed in the string color.
- The lines after the blank line of the string enclosed in double quotes are not displayed in the string color.

Current version:
current

Fixed version:
fixed


sdottaka:
Thank you!

PR #835, opened 2021-06-21
Add silent option to RegSvr32

antoine-aumjaud:
To avoid useless popin, add the option silent to regserv32


sdottaka:
Thanks for the PR.

I don't want to add the /s option in a fixed way, because we won't notice if the registration fails. Also, if you are going to fix it, you should fix Register.bat as well.

If the /s option is specified in RegisterPerUser.bat and Register.bat, I have modified it in the following commit to make it a silent registration.

https://github.com/WinMerge/winmerge/commit/1ad63ce772a4829dd3134f9f225a1f6c4d993611

antoine-aumjaud:
Thanks for these modifications!

PR #857, opened 2021-07-04
Fix an issue where a file is deselected when returning to the folder compare window after opening the file compare window by double-clicking the file in the folder compare window.

tjmprm77:
Sometimes a file is deselected when returning to the folder compare window after opening the file compare window by double-clicking the file in the folder compare window.

For example, it occurs by the following procedure.
Step 0. Set as follows.
  - The "Include subfolders" option is on.
  - The display of the message box "The selected files are identical." is enabled.

Step 1. Compare directories with the same three files.
Step 2. Adjust the window size to just fit the three files.
Step 3. Double-click on the third file. This open a file compare window with the message "The selected files are identical."
Step 4. Close the message box and select the folder compare window. At this time, the third file is deselected.
diff

It is a bit inconvenient because we lose track of which file had been checked when checking the differences between the files in the folder one by one, if the file is deselected when we return to the folder compare window. 

Regarding this behavior, I suspect it is not good that CListView::OnLButtonDblClk() is executed while the focus is out of the folder compare window by opening the file compare window.

With this PR, CDirView::OnLButtonDblClk() now executes CListView::OnLButtonDblClk() only when it is in focus.

### DIFF
--- a/Src/DirView.cpp
+++ b/Src/DirView.cpp
@@ -491,7 +491,8 @@ void CDirView::OnLButtonDblClk(UINT nFlags, CPoint point)
 			OpenSelection();
 		}
 	}
-	CListView::OnLButtonDblClk(nFlags, point);
+	if (GetFocus() == this)
+		CListView::OnLButtonDblClk(nFlags, point);
 }
 
 /**


sdottaka:
Thank you!

PR #882, opened 2021-07-17
Fix the Select Files or Folders dialog.

tjmprm77:
Fix an issue where the enabled state of "Folder: Filter" and "File: Unpacker Plugin" is not updated when loading a project with the "Load Project..." drop down button.


sdottaka:
Thank you!

PR #892, opened 2021-07-22
Fix the Select Files or Folders dialog.

tjmprm77:
Fix an issue where plugin items are added to the "Save Project ..." drop down button when files (instead of directories) are specified in the "File or Folder" fields.


sdottaka:
Thank you!

Issue #489, opened 2020-11-25
Save compare options in project file


bkisdi:
The `Save Project...` operation should save the compare options set by the top right `Options...` button into the project file (`.WinMerge`) - and the  `Open Project...` should restore them, of course.

(This is a feature request, but I can't label it :(. )

ranjwarrior:
Looking for additional comments/clarifications - Based on my understanding of current WinMerge behavior, this is how most IDEs/Editors work. Project settings are saved separately from global settings. Some issues in combining both and implementing this feature are:
1. When opening multiple project files global settings will be overwritten by the last opened project
2. If an existing compare was completed with current global settings and a new project that overwrites current settings are opened, user will not know options used in the first compare

If this is to be implemented, should the project file (xml format) and the settings file (ini format) be combined into xml format when saving? Or, these can be saved separately together into a zip format (with the xml and ini files independently) and upon open, imported separately.

sdottaka:
Implemented by PR #915

PR #915, opened 2021-08-13
Add a feature to save/restore compare options to/from a project file. (#489)

tjmprm77:
Add a feature to save/restore compare options to/from a project file requested in issue #489.

In this pull request, I changed the behavior of the "Options..." popup menu in the Select Files or Folders dialog. Is that okay?
The reason for this change is as follows.
"Include Subfolders" and file filters are not saved in the WinMerge option when we have loaded the project or change values on the dialog, but are saved in the WinMerge option when we then click the "Compare" button to start the comparison.
I think it's better that the compare options are also saved in the WinMerge option with the same scheme as those.
So I changed the behavior of the "Options" popup menu as follows:
- Even if the values in the popup menu are changed, it is not saved in the WinMerge option at that time.
- The values in the popup menu are saved in the WinMerge options when the "Compare" button is clicked.

Also, due to this change, the "Include Subfolders" item in the popup menu has exactly the same function as the check box and is no longer needed, so it has been deleted.


sdottaka:
Thank you!

>Is that okay?

LGTM. But, I might change GetOptionsMgr()->SaveOption() in the CMergeApp::LoadAndOpenProjectFile() function to GetOptionsMgr()->Set() later. This is to make the change in settings temporary, similar to the behavior of command line options such as /ignorews and /ignorecase.

tjmprm77:
Thank you for your reply.
I am relieved that there seemed to be no problem.

> But, I might change GetOptionsMgr()->SaveOption() in the CMergeApp::LoadAndOpenProjectFile() function to GetOptionsMgr()->Set() later. This is to make the change in settings temporary, similar to the behavior of command line options such as /ignorews and /ignorecase.

Sure, as you say, I think it might be better to apply the compare options as temporary settings when opening the project file, instead of saving them to the WinMerge options.

BTW, although it may have something to do with the above topic, file filters, plugins and comparison options are now always saved/restored to and from the project file, but I think it's useful if users can choose whether or not to save/restore them. 
If you are in favor of adding this feature, I will implement it and send a PR.

sdottaka:
> If you are in favor of adding this feature, I will implement it and send a PR.
I agree with the implementation of that feature.

tjmprm77:
Thank you.
I will start implementing it.

PR #953, opened 2021-09-05
Add a feature to set items saved to or restored from the project file.

tjmprm77:
Add a feature to set items saved to or restored from the project file proposed in PR #915.
project


sdottaka:
Thank you!

PR #966, opened 2021-09-13
compiler-calculated maximum value for `m_SourceDefs`

GitMensch:
removing the need to manually adjust the size when `TextType` is adjusted


sdottaka:
Thank you!

GitMensch:
You're welcome. I guess I may add the COBOL language syntax sooner than later, too. The recent Smarty addition, together with the existing syntax definitions look like a good start, but I obviously need to put some more testing (and up-front setup of the dev environment) into that - so not too soon...

PR #974, opened 2021-09-14
Updated Swedish translations

timint:
The previous translations were absolutely terrible. I mean literally terrible.
I have redone it completely. The previous has words that were invented, weird upper- amd lowercases, words nobody uses. And sometimes just jibberish.


sdottaka:
Thank you!

The names of past translators have disappeared, so I'll put them back.

timint:
Somone should send them back to school rather than giving them credits. 😆 That was a complete wreck of a translation. The worst I have ever seen (not exaggerating).

PR #980, opened 2021-09-21
2.16.15 Turkish translation update  

kayazeren:
2.16.15 Turkish translation update


sdottaka:
Thank you!

 For Translations/ShellExtension/Turkish.po, Encoding has been changed from UTF-8 to CP1254, but I want all languages to be UTF-8, so I will change it back to UTF-8.

PR #987, opened 2021-09-27
Loop counter should be the same type as the count type.

sredna:
It is unlikely that somebody selects more than 0xffff files but just in case, don't loop forever.


sdottaka:
Thank you!

PR #995, opened 2021-10-03
Fix an issue where filters are not applied correctly when opening a project file containing multiple items with different filters.

tjmprm77:
In the current version, filters are not applied correctly when opening a project file that contains multiple items with different filters, such as the attachment below.
[filter.zip](https://github.com/WinMerge/winmerge/files/7272813/filter.zip)

Specifically, as shown in the image below, for the first project item, the filter is displayed in the status bar correctly, but the files displayed are ones filtered by the filter of the second project item.
filter

The processing related to the filter when opening the project is as follows.
Step 1. Set the filter described in the project file to the global file filter. (CMergeApp::LoadAndOpenProjectFile(), in the main thread)
Step 2. Stores a pointer to a global file filter in the diff context of the CDirDoc class. (CDirDoc::InitDiffContext(), in the main thread)
Step 3. Generate an item collection thread (DiffThreadCollect()) and a folder compare thread (DiffThreadCompare()), pass the diff Context and start them. (CDirDoc::Rescan(), in the main thread)
Step 4. List directories and files using the passed diff context. (in the item collection thread and the folder compare thread)

In the process of opening the attached project file, before the item collection thread and the folder compare thread use the file filter in the diff context in Step 4 for the first project item, the main thread has already completed Step 1 for the second project file. As a result, the file filter in the diff context has been changed to the filter of the second project item.

For this reason, I modified the process to duplicate the global file filter and set it in the file filter of the diff context of the CDirDoc class, instead of using the pointer to the global file filter as is.
(I just wanted to duplicate the global file filter, but the fix was a bit big.)

### DIFF
--- a/Src/DirDoc.cpp
+++ b/Src/DirDoc.cpp
@@ -253,7 +253,8 @@ void CDirDoc::InitDiffContext(CDiffContext *pCtxt)
 	// Make sure filters are up-to-date
 	auto* pGlobalFileFilter = theApp.GetGlobalFileFilter();
 	pGlobalFileFilter->ReloadUpdatedFilters();
-	pCtxt->m_piFilterGlobal = pGlobalFileFilter;
+	m_fileHelper.CloneFrom(pGlobalFileFilter);
+	pCtxt->m_piFilterGlobal = &m_fileHelper;
 	
 	// All plugin management is done by our plugin manager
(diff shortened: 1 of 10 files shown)


sdottaka:
Thank you!

PR #1038, opened 2021-11-03
Add the feature to display tooltips on the MDI tab.

tjmprm77:
Show tooltips when the title is too long to display the entire title.
tooltips


sdottaka:
Thank you!

PR #1055, opened 2021-11-23
Fix PHP syntax highlighting.

tjmprm77:
Fix an issue where text is not highlighted correctly if there is no space or symbol between a single line comment or variable and the end tag "?>".

Current version:
curent

Fixed version:
fixed


sdottaka:
Thank you!

Issue #460, opened 2020-10-17
Do not save compare options changed on the fly.


SamuelPlentz:
I usually want to compare everything:
- all whitespace differences
- all case sensitive differences
- all empty line differences
- all linebreak differences
- all codepage differences
- all comment differences

But in special cases I have to deactivate 1 or 2 rules (especially "whitespace" and "case sensitive") of that list. If I do this and restart Winmerge, the rules are still deactivated and I probably miss important differences.

I want to reset my compare options on each restart of Winmerge to my defaults. IMO other options are not affected.

What are your ideas to handle this use case? Is it even possible yet?

I propose to add an option and a button:
- [x] Save compare options automatically

`Save compare options now`

Implementing note: Some options like "compare whitespace" can be changed on the fly. They need to be reverted as well.

SamuelPlentz:
@sdottaka: I consider to create a pull request for this, but I don't want to start, before you comment on this.
I am open for other ideas facing that issue.

sdottaka:
I agree with your idea, but I prefer all options, not just the comparison option.

FYI, you can call CRegOptionsMgr::SetSerializeing(false) to prevent the settings from being saved in the registry.

SamuelPlentz:
I made a mockup, how it could look:

![Winmerge](https://user-images.githubusercontent.com/1890854/96452142-e1199f80-1218-11eb-8333-555065408646.png)

Colors and strings could be adjusted. I think it is a good idea to provide some kind of warning banner.
Do you agree with the position (section `General` below the `Language` entry) and the the additional banner?

I'am not sure about the strings.

Instead of:
`The option "save options automatically" is disabled. Changes are lost on restart.`

It could be:
`As "save options automatically" is disabled, changes are lost on restart.`

SamuelPlentz:
Probably the warning banner should be yellow instead of red, because it works like intended.

sdottaka:
> Do you agree with the position (section General below the Language entry) and the additional banner?
I agree with you.
If the implementation is cumbersome, you can implement it by displaying a message box instead.

mrLithe:
I propose to make buttons presets on the toolbar for different options of user settings.

SamuelPlentz:
Mh, my motivation to implement this is somewhat reduced, as I found a neat workaround suitable for me.

[...]

PR #1071, opened 2021-12-06
Add an "Defaults" section to the ini file

SamuelPlentz:
After reading the "WinMerge" section try to read the "Defaults" section.
Overwrite existing entries in "iniFileKeyValues" with the ones from the "Defaults" section.
This section is not written to, as it should contain default entries, that are reverted at each restart of WinMerge.
More informations see #460.


sdottaka:
Thank you!

PR #1120, opened 2022-01-09
Fix an issue where WinMerge sometimes crashes when executing "Refresh Selected" in the folder compare window.

tjmprm77:
In the current version, WinMerge sometimes crashes when executing "Refresh Selected" in the folder compare window.
In my environment, there is a high probability of a crash, especially if the "Include Subfolders" option is on and many items are selected.
For example, the situation is as shown in the image below. The set of directories compared at this time is attached.
[RefreshSelected.zip](https://github.com/WinMerge/winmerge/files/7834709/RefreshSelected.zip)
RefreshSelected

As a result of monitoring this phenomenon, the situation when "Refresh Selected" is executed is as follows.

- The collection thread DiffThreadCollect() sends the UI update event CDiffThread::EVENT_COLLECT_COMPLETED as soon as it starts because there is nothing to do.
- In the compare thread, DIFFITEM::RemoveChildren() called from DirScan_UpdateMarkedItems() removes DiffItems.
- In response to the UI update event CDiffThread::EVENT_COLLECT_COMPLETED, the UI update process such as CDirView::OnUpdateUIMessage () and CDirView::ReflectGetdispinfo() runs on the main thread.
  At this time this process may refer to the deleted DiffItem in the compare thread running in parallel.

This PR fixes this issue by performing the UI update process after the process that may delete DiffItems is completed.

### DIFF
--- a/Src/DiffThread.cpp
+++ b/Src/DiffThread.cpp
@@ -73,7 +73,7 @@ bool CDiffThread::ShouldAbort() const
  */
 unsigned CDiffThread::CompareDirectories()
 {
-	assert(m_pDiffParm->nThreadState != THREAD_COMPARING);
+	assert(m_pDiffParm->nThreadState != THREAD_COMPARING && m_pDiffParm->nCollectThreadState != THREAD_COMPARING);
 
 	m_pDiffParm->context = m_pDiffContext;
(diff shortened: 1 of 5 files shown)


sdottaka:
Thank you!

PR #1147, opened 2022-01-22
Clipboard Compare

sdottaka:
This PR adds the ability to compare the two most recent contents of the clipboard history.
This feature is available on Windows 10 version 1809 or higher and WinMerge 64-bit version.

Select File → Open Clipboard menu item to compare the clipboard contents.
You can also compare by pressing Ctrl+V when the child MDI window is not visible.

clipboard-history


Andrej730:
Just found it by accident by hitting Ctrl-V in a new WinMerge Window - thank you so much for this feature, it's very convenient! This is the first time I'm seeing a software utilizing Windows clipboard history, now I see the potential how it can also be utilized in other apps too.

PR #1164, opened 2022-01-27
Update Russian.po

wvxwxvw:


wvxwxvw:
In Russian, the search is case-sensitive regardless of the flag.

sdottaka:
Thank you!

> In Russian, the search is case-sensitive regardless of the flag.

Unfortunately, WinMerge's case-insensitive search only supports ASCII characters.

PR #1188, opened 2022-02-10
Update Corsican translations and change encoding

Patriccollu:
Hello,

This is a small update on two members `ReadMe-Corsican` and `Corsican.isl`.

The encoding was automatically changed from windows-1252 to UTF-8.

Best regards,
Patriccollu.


sdottaka:
Thanks for the PR.
For Windows XP support, an older version of InnoSetup is used. Since this version cannot use UTF-8, I will revert the encoding later.

PR #1189, opened 2022-02-12
Fix an issue where comparison results are not displayed correctly when "Refresh Selected" is performed by selecting an item that has a directory and file with the same name.

tjmprm77:
Comparison results are not displayed correctly when "Refresh Selected" is performed by selecting an item that has a directory and file with the same name as shown below.

Before executing "Refresh Selected":
01_before

After executing "Refresh Selected" for the directory:
02_after_dir

After executing "Refresh Selected" for the file:
03_after_file

This PR fixes this issue by changing to judge that a directory or file exists only if the rescanned item type (directory or file) matches the original diffItem type in the diffItem update process.

### DIFF
--- a/Src/DirItem.cpp
+++ b/Src/DirItem.cpp
@@ -98,3 +98,12 @@ void DirItem::ClearPartial()
 	size = DirItem::FILE_SIZE_NONE;
 	flags.reset();
 }
+
+/**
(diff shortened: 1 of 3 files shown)


sdottaka:
Thank you!

Issue #1008, opened 2021-10-13
Semi matching lines are not aligned vertically


mdonatas:
![image](https://user-images.githubusercontent.com/483659/137171756-fdf66e01-30aa-4021-8bb0-15e84e0920c7.png)

It seems WinMerge matches the lines that differ and highlights minor differences but for some reason those lines are not aligned vertically which makes this diff much harder to read than if those lines were matched vertically.

left
```
curl -L https://www.myget.org/F/blah/auth/b477hrd-cdd-k-ey-blah/api/v3/flatcontainer/nupkg --output deployer.zip
unzip -q deployer.zip -d deployer

```

right
```
MYGET_KEY="$(echo '%MYGET_API_KEY%' | sed 's/credentialsJSON:\(.*\)/\1/')"

curl -L "https://www.myget.org/F/blah/auth/$MYGET_KEY/api/v3/flatcontainer/nupkg" --output "deployer.zip"
unzip -q -o deployer.zip -d deployer

```

PR #1200, opened 2022-02-18
Improve vertical alignment string

Gitoffthelawn:
I experienced the same issue as in: https://github.com/WinMerge/winmerge/issues/1008

Both that user and I were misled by the "Match similar lines" string.  In English, it's confusing within the context of a comparison application because "match" is also a comparison term.

As such, I improved it to be "Align similar items".


Gitoffthelawn:
You're welcome.  Thank *you* for the helpful tool!

PR #1246, opened 2022-03-13
Fix renaming process in folder compare window.

tjmprm77:
The comparison result is not displayed correctly when renaming a file (directory) that exists on only one side to a name where the directory (file) with the same name exists.

Example 1. Rename file AAA to BBB when file AAA (left only) and directory BBB (right only) exist.
e1

Example 2. Rename directory BBB to AAA when file AAA (left only) and directory BBB (right only) exist.
e2

This PR fixes this issue by changing to judge that a directory or file exists only if the type (directory or file) of renamed item matches the original diffItem type in the diffItem update process.


sdottaka:
Thank you!

PR #1258, opened 2022-03-18
[Big PR - big changes] A lot of refactor and optimization commits

GermanAizek:
@kimmov, @sdottaka
check it and feedback please.


sdottaka:
Thank you for the PR!
As for the MergeDoc.cpp fixes in commit 2bc5799, I will revert some of them later, as the scope optimization has broken the process of restoring saved states.

PR #1299, opened 2022-04-23
Fix an issue where items with different case are not displayed correctly in the folder compare window when comparing three directories.

tjmprm77:
Both file names are displayed for items with different case in the folder comparison window when comparing two directories.
(For example, "TEST.TXT | test.txt")
In the current version, items are not displayed correctly when comparing three directories.


sdottaka:
Thank you!

PR #1319, opened 2022-05-02
Fix an issue where WinMerge crashes depending on the filename when generating a file compare report.

tjmprm77:
WinMerge crashes depending on the filename when generating a file compare report.
(For example,"&&")
This is because the buffer of the replaced string is not properly resized in the process of replacing the special characters contained in the file name with the HTML entities.


sdottaka:
Thank you!

PR #1324, opened 2022-05-05
Fix report generation process.

tjmprm77:
Fix an issue where the middle directory is not output in the title of the directory compare report for three directories.


sdottaka:
Thank you!

PR #1333, opened 2022-05-13
Modify "Tools - Generate Report" (Add column width definition to html output)

stonee-K:
Add column width definition to html output

GenerateReport


sdottaka:
Thank you for the PR!

I am a little concerned that the margin width is fixed even when line numbers are not displayed, so I may fix this later.

PR #1338, opened 2022-05-16
Update Japanese.po (Codepage dialog)

stonee-K:
"both" means loading and saving, not left pane and right pane.

fileencodingdialog


sdottaka:
Thank you!

PR #1342, opened 2022-05-19
Fix typo: GPL

stonee-K:
diffutils ver 2.9

> 2007-07-19
> * src/ifdef.c: Fix typo: "GNU DIFF General Public License" should be
> "GNU General Public License".  Reported by Erich Guenther.
> * src/side.c: Likewise.


sdottaka:
Thank you!

PR #1344, opened 2022-05-20
Fix report generation process about replacement with HTML entity

tjmprm77:
The string with the hyperlink to the file compare report is not replaced with the HTML entity, when generating an HTML-style folder compare report with the "Include File Compare Report" option enabled.
Therefore, for example, if the filename field is "TEST.TXT|test.txt|\<None\>", the "\<None\>" part will not be displayed properly when the report is displayed in the browser.


sdottaka:
Thank you!

PR #1351, opened 2022-05-30
Use isupper+tolower instead of islower+toupper (diffutils ver2.7)

stonee-K:
diffutils ver2.7
> Mon Sep 12 15:52:22 1994
> Use isupper+tolower instead of islower+toupper, for consistency.


sdottaka:
Thank you!

PR #1358, opened 2022-06-04
Make the file name refreshed to detect case changes when running "Refresh Selected".

tjmprm77:
In the current version, the file name display is not updated when we change the case of the file name and then execute "Refresh Selected".
For example, even if we rename "TEST.TXT" to "test.txt" and then perform "Refresh Selected", the item display remains "TEST.TXT".
In this PR, the file name is now updated when we run "Refresh Selected".


sdottaka:
Thank you!

PR #1360, opened 2022-06-06
Initialize variables defined at "diff.h"

stonee-K:


sdottaka:
Thank you for the PR.

Is this PR to solve some kind of problem?

They are all global variables, so they are originally initialized with 0.
For line_end_char, '\n' is set in the constructor of CDiffWrapper.

PR #1372, opened 2022-06-19
Fix an issue where paths with different case are not displayed correctly in the folder column of the folder compare window when comparing three directories.

tjmprm77:
Both paths are displayed for items that have different case paths in the folder column of the folder compare window when comparing two directories.
In the current version, paths are not displayed correctly when comparing three directories.


sdottaka:
Thank you!

PR #1374, opened 2022-06-20
Modify textbox behavior (CompareMSExcelFiles options dialog)

stonee-K:
* Disable Shift + Numeric Key
 * Enable CTRL-X, CTRL-C, CTRL-V, CTRL-Z
 * Allow pasting when the clipboard string is numeri
 
excel_plugin


sdottaka:
Thank you!

PR #1377, opened 2022-06-22
Keep hidden items

mauriciogracia:
This is a PR in progress, looking for some previous validation before continuing 

The list of hidden items is kept and updated every time a new item is hidden, or if all hidden items are shown again

I'm currently using a new xml node of these hidden items
image

When opening the project, that list will be read and folders/files will be hidden to keep context

Also when closing the project with hidden-items the Save method will be triggered.

Not sure if this functionality needs to be behind a flag/option/setting.


mauriciogracia:
1. The <hidden-list> is now kept under <paths>
2. I have change to use the Allman identation style
3. I have updated the README.md 

Still not sure if this functionality needs to be behind a flag/option/setting.

sdottaka:
The following Options dialog allows you to set which items are handled when opening, loading, and saving a project.
It would be helpful to be able to set the hidden items as well.

image

mauriciogracia:
I still have not implemented the Project/Options that you mention above....

However I've managed to read the hidden-items from the projectFile (.winmerge) and I need to pass it to the dirview so they can be hidden but I'm a bit lost about how to do that (similar to the way I modified the SaveFileProject method)

Below is a test with some hardcoded items, could you guide me into this part please ?
image

sdottaka:
How about doing the following?

1. Create a class that inherits from the `OpenFileParams` class(MainFrm.h) so that you can store hidden items in that class.
2. Get the hidden items from `projItem` in `CMergeApp::LoadAndOpenProjectFile()` and set it to the instance of 1.
3. Pass the instance of 1. as an argument to `GetMainFrame()->DoFileOrFolderOpen()` in `CMergeApp::LoadAndOpenProjectFile()`.
4. In `CMainFrame::DoFileOrFolderOpen()`, take the hidden items from the instance of 1. and set the hidden items to `pDirDoc-> GetDiffContext()->m_vCurrentlyHiddenItems` near line 1380.
5. Remove `//` in `// HideItems (GetDiffContext (). M_vCurrentlyHiddenItems);` in DirView.cpp

mauriciogracia:
I read your comment many times, and did not get pass step 2

My original approach was to have two lists of hidden-items: 
- one that is kept each time an item is hidden `CDiffContext.m_vCurrentlyHiddenItems` and that saved to the .winmerge file
- the other one that is loaded from the .winmerge file `ProjectFileItem.m_vSavedHiddenItems`. 

I thought they needed to be different lists, but now that I think about it it might not be necesary anymore 

Anyhow currently: 
- I'm saving the hidden items using the `saveHiddenItems` from `ProjectFile.cpp`
- I'm reading the hidden items with a change I did on method `void characters(const XMLChar ch[], int start, int length)` also in `ProjectFile.cpp` 

I was thinking about keeping a single list of HiddenItems in property of `ProjectFile` or `ProjectFileItem`, since saving and reading is happening there and I have already added logic ther, but from DirView (where the events of item being hidden happen) I dont seem to have access to the currec ProjectItem*

In your explanation you have not mention if I need to change the way the data is being saved, my guess is that I need to a find a class that is common to both the save and load of a project and store a singe list of hidden items there, so that when loading I will set the items to that property, and will keep it updated as I do know

**Hope any of the above is clear, this is my first C++ project, and navigating the code has been quite difficult, maybe you could help me with a commit on my branch to wrap this up ?**

mauriciogracia:
@sdottaka I have merged you PR, it works great thanks, will work on the save/load options for the save items and keep you posted - thanks for all your help and patience

mauriciogracia:
I have added the "Hidde Items" to the Project/Options as agreed
image

It would be nice to avoid the scroll by making the vertical space bigger but not sure if that could affect anything below (that is currently not visible)

Docs/Users/ChangeLog.md
```diff
@@ -562,11 +562,13 @@
 
 - Add a feature to save/restore compare options to/from a project file.(#498)
     (PR #915)
+- Add a feature to save/restore hidden items to/from a project file.(PR #1377)
```
sdottaka: Since here is a ChangeLog for version 2.16.15, which has already been released, it must be included in the ChangeLog for 2.16.22, which will be released in the future.

Docs/Users/ChangeLog.md
```diff
@@ -562,11 +562,13 @@
...
 
 ### Options dialog
 
 - Add a feature to set items saved to or restored from the project file.
     (PR #953)
+- New setting to decide when to save/restore hidden items when project is saved/loaded (PR #1377)
```
sdottaka: Same as above

README.md
```diff
@@ -55,10 +55,12 @@ git clone https://github.com/winmerge/winmerge
 cd winmerge
 git submodule init
 git submodule update
-DownloadDeps.cmd
+>DownloadDeps.cmd
```
sdottaka: I want to delete ">" because it causes an error when I paste it into the command prompt.

README.md
```diff
@@ -55,10 +55,12 @@ git clone https://github.com/winmerge/winmerge
...
 BuildAll.vs2022.cmd [x86|x64|ARM|ARM64] or BuildAll.vs2019.cmd [x86|x64|ARM|ARM64] or BuildAll.vs2017.cmd [x86|x64|ARM|ARM64]
 ~~~
 
+Keep in mind that the `DownloadDeps.cmd` needs to be run from the VS Developer Command Prompt
```
sdottaka: I would like to be able to run these commands on cmd.exe and not VS Developer Command Prompt.

On my PC DownloadDeps.cmd works fine on cmd.exe, but on your PC did DownloadDeps.cmd only work with VS Developer Command Prompt?

Src/OptionsProject.h
```diff
@@ -19,9 +19,10 @@ enum class Item
 	FileFilter,
 	IncludeSubfolders,
 	UnpackerPlugin,
-	CompareOptions
+	CompareOptions,
+	HiddenItems
```
sdottaka, line 23: Line 15 of OptionsProject. cpp also needs to be modified as follows:
~~~
const String ItemKeyword[ItemCount] = { _T("FileFilter"), _T("IncludeSubfolders"), _T("UnpackerPlugin"), _T("CompareOptions") };
~~~
↓
~~~
const String ItemKeyword[ItemCount] = { _T("FileFilter"), _T("IncludeSubfolders"), _T("UnpackerPlugin"), _T("CompareOptions"), _T("HiddenItems") };
~~~


Also, it is great that with this addition, Hidden Items appear in the Options dialog and the check state is saved and loaded. However, there appears to be no code to reflect this setting in the saving and loading of the project.
I would appreciate it if you could implement this with reference to the implementation of the other items.
A search for `Options::Project::Get` or `projItem.SetSave` would give you some hints.

Src/DirView.cpp
```diff
@@ -3264,14 +3265,61 @@ void CDirView::OnUpdateItemRename(CCmdUI* pCmdUI)
...
+bool CDirView::IsItemToHide(const String& currentItem, const std::vector<String>& ItemsToHide) const
+{
+	return std::find(ItemsToHide.begin(), ItemsToHide.end(), currentItem) != ItemsToHide.end();
+}
+
+/**
+ * @brief hides items specified in the .winmerge file
+ */
+
+void CDirView::HideItems(const std::vector<String>& ItemsToHide)
```
sdottaka, line 3299: When TreeMode is enabled, hiding a file in a subfolder and saving the item in a project file works correctly, but loading the project file did not hide the file.
This is because the HideItems function only handles the items currently visible in the list view.
I would appreciate it if you could implement it in a way similar to the ExpandAllSubdirs function of DirActions.cpp.

sdottaka (changes requested):

README.md
mauriciogracia: To be honest now I doubt which one I had issues running from I will keep this part of the instructions as they were

README.md
mauriciogracia: Done (that was like an * for the comment below) I have remove it

Docs/Users/ChangeLog.md
mauriciogracia: So this is done after is merged ?

Docs/Users/ChangeLog.md
mauriciogracia: this is done after is merged ?

sdottaka: Please move the line you added to the top of the ChangeLog.md file as follows or delete it.

If you delete it, I will add the line to ChangeLog.md after the merge or just before the version 2.16.22 release.

~~~

## WinMerge 2.16.22 - 2022-07-27

### Project file

- Add a feature to save/restore hidden items to/from a project file.(PR #1377)

### Options dialog

- New setting to decide when to save/restore hidden items when project is saved/loaded (PR #1377)

~~~

In this project, the addition to ChangeLog.md is usually done by the maintainer, but it would be helpful if the person making the pull request could add it, as it would save time.

Src/DirView.cpp
mauriciogracia: I have tried expanding the folders while searching for the item to hide, but have not tested for reason that I mention in other comment

Src/OptionsProject.h
mauriciogracia: I have tried to follow the same pattern, but now it seems that no matter what I check on the check boxes the hidden-items are not saved. This has stopped me from testing the "TreeView" behavior that you mentioned (below)

Branch has been updated with most of the changes you have requested

sdottaka: Please see https://github.com/mauriciogracia/winmerge/pull/2

mauriciogracia:
Thanks for the PR (changes) I have merged and is working fine, also updated the Changelog.md as requested

1. Does WinMerge keep track of "unsaved changes" to prompt the user for "There are unsaved changes in your project, do you want to save them before closing it?" (if not, could that be an improvement to make?)
2. Did you notice that I added many translations texts to spanish in the translation folder ? can I continue translating the rest of the texts maybe on a different PR ?
3. I have notice that the option "Save Project" exists but instead of prompting for confirmation to save, it takes you to window where you can Save, Load/Open, which for me is a bit confusing (by the way what is difference between load/open)
image

All of the above comments are made from the respect and love that I have for this project and if you consider that any of the above suggestions are good improvements let me know to start working on them.

Please review my latest changes I hope that we can get this PR merged

sdottaka:
Thank you for updating the PR. This PR has been merged.

1. Unfortunately, the current WinMerge does not display such a message.
    It's a good idea to display such a message, but only if you open it from a project file.

2.  Yes, I noticed it. The update of Spanish.po has been stopped, so we welcome PR.

3. I think it's a good idea to be able to save project files without displaying the "Select files or folder" window.
     "Load" loads the contents of the project file and reflects it in the "Select files or folders" window, no comparison. "Open" reads the contents of the project file and compares the files or folders.

mauriciogracia:
I have created the local branch for the Spanish translation, and will keep you posted.

Thanks for all your help, this could become a nice hobby.

mauriciogracia:
> 3\. I think it's a good idea to be able to save project files without displaying the "Select files or folder" window.
>         "Load" loads the contents of the project file and reflects it in the "Select files or folders" window, no comparison. "Open" reads the contents of the project file and compares the files or folders.

What you mention on item 3. feels like it could be just a LOAD option, that has a checkbox that says "compare after load" (not sure if its the only difference

PR #1392, opened 2022-07-03
Fix renaming process in folder compare window.

tjmprm77:
The comparison result is not displayed correctly when renaming a file (directory) that exists on only one side to a name where the file (directory) with the same name exists.
This PR fixes this issue by doing the following:
- Rescan the renamed item if the side flag changes due to renaming.
- Delete the item with the same file name as after renaming.

(PR #1246 dealt with cases where a file and a directory have the same name, and this PR deals with cases where two or three files have the same name or two or three directories have the same name.)

### DIFF
--- a/Src/DirView.cpp
+++ b/Src/DirView.cpp
@@ -48,6 +48,7 @@
 #include "PatchTool.h"
 #include "SyntaxColors.h"
 #include "Shell.h"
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you!

PR #1405, opened 2022-07-17
Fix: "Character level" option is always ignored

stonee-K:
char_level_diff

char_level_diff_2


sdottaka:
Thank you for the PR.
Indeed, the current implementation is not a true character-by-character comparison, so I think it is very good to be able to do true character-by-character comparison.

However, the reason for the current implementation (word-by-word comparison followed by character-by-character range adjustment) is performance. (It may not be a problem in CPU speed today.)

So can we leave the default as the current implementation and make this true character-by-character comparison an option?

This PR also causes UnitTests to fail.
I would also appreciate it if you could fix UnitTests.

image

PR #1408, opened 2022-07-18
Fix an issue where the folder column is not updated for child items after renaming a directory in the folder compare window.

tjmprm77:
The folder column is not updated for child items after renaming a directory in the folder compare window.
This PR fixes this issue by updating the path of the child items after renaming the directory.
This PR also fixes an issue where file comparisons of child items after renaming a directory would result in empty files comparison instead of the intended files comparison.


sdottaka:
Thank you!

PR #1430, opened 2022-07-28
Create Midnight.ini

HaroldPetersInskipp:
Added an additional color scheme that is easy on the eyes.

### DIFF
--- a/ColorSchemes/Dark Mode.ini
+++ b/ColorSchemes/Dark Mode.ini
@@ -1,0 +1,159 @@
+; Dark Mode color scheme for WinMerge
+; based on https://github.com/WinMerge/winmerge/blob/master/ColorSchemes/Solarized%20Dark.ini
+; license: MIT
+[WinMerge]
+; base03
+Custom Colors/0=0x362b00
+; base02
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you for the PR.

The name "Dark Mode" looks strange compared to the other schema names, so could you please remove the "Mode" and name it something else?

Also, as you can see below, the colors of the folder window and other colors have not changed much from Solarized Dark. Can you make them a little different from Solarized Dark?


image

HaroldPetersInskipp:
I have updated and renamed the color scheme to better differentiate it from Solarized Dark.

sdottaka:
Thank you for updading the file.

The diff items in the folder window are a little hard to see for me, is this OK for you?

image

HaroldPetersInskipp:
I've altered the value of the background for diff items in the folder window to better contrast with the text. I didn't want increase the contrast too high here, as that would defeat the purpose of a dark theme, but I think this is a good compromise between ease of readability and lack of eye strain.

sdottaka:
Thank you!

PR #1434, opened 2022-07-31
Disable rename operations when in read-only mode in the folder compare window.

tjmprm77:
Disable rename operations when in read-only mode in the folder compare window.


sdottaka:
Thank you!

PR #1451, opened 2022-08-13
Fix an issue where renaming to a file name or directory name containing "\" or "/" is not done properly.

tjmprm77:
In the current version, "\" and "/" can be used when renaming files and directories in the folder compare window, and the behavior is as follows.

1. Rename AAA.txt to BBB\CCC.txt.
   In this case, the item displayed in WinMerge is BBB\CCC.txt, but actually AAA.txt is renamed to CCC.txt and a separate directory BBB is created.

2. Rename AAA.txt to BBB/CCC.txt.
   In this case, the item displayed in WinMerge is BBB/CCC.txt, but actually AAA.txt is renamed to CCC.txt

This PR adds a process to check if the specified name is valid when doing a rename, and prevents the use of "\" and "/".


sdottaka:
Thank you!

PR #1459, opened 2022-08-16
Update Brazilian.po

Marcello-mco:
Still indicating one translation missing. This time I've downloaded PoEdit and used it to translate. Let's check the result.


sdottaka:
Thank you!

PR #1463, opened 2022-08-19
Update Brazilian.po

Marcello-mco:
My last attempt in completing 100% translation.

### DIFF
--- a/Translations/WinMerge/Brazilian.po
+++ b/Translations/WinMerge/Brazilian.po
@@ -1,4364 +1,3848 @@
 # This file is part from WinMerge <https://winmerge.org/>
 # Released under the "GNU General Public License"
 #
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you for the PR. But this PR seems to be broken as follows

image

Marcello-mco:
> Thank you for the PR. But this PR seems to be broken as follows
> 
> image

Don't know what to say, it seems correct from here. But I've updated the file again. I've joined the split lines (msgid "" followed by lines below) and inserted translation for "The file or folder name is invalid.")

sdottaka:
Thank you for re-uploading. I have merged this PR.
You probably uploaded the contents of the English.pot file, not the Brazilian.po file.

PR #1466, opened 2022-08-23
Fix an issue where filenames containing "&" are not displayed properly in the MDI tab bar and its tooltips.

tjmprm77:
Filenames containing "&" are not displayed properly in the MDI tab bar and its tooltips.
For example, comparing "AAA&BBB.txt" and "CCC&DDD.txt" is as follows.
tab
tooltips

This PR fixes this issue by doing the following:
- Add processing to escape "&" in the MDI tab bar.
- Add TTS_NOPREFIX to the tooltip style.


sdottaka:
Thank you!

PR #1481, opened 2022-09-11
The feature will allow the user to right-click the selected lines and…

srgank:
The feature will allow the user to right-click the selected lines and… add them to Line Filter so that those lines added to the Line Filter will be ignored if found in any file.

https://user-images.githubusercontent.com/7500655/189524283-53b6790e-8572-41a5-a45b-0235c9915f68.mp4


sdottaka:
Thank you for your PR. I'll take a look.

sdottaka:
Thank you for your PR.
I merged this PR.
I would like to make the following modifications later.
- Delete unnecessary blank lines etc. added
- The mnemonic (&F) of the "Add to Line &Filters" menu item conflicts with the "Add this change to Substitution &Filters" menu item.
- Since the line filter is a regular expression, the string added to the line filter must be converted to a string that is valid as a regular expression.
- The HideFilterLines method doesn't actually hide the lines, so the name feels inappropriate
- Line filters are regular expressions, so the HideFilterLines method must search with regular expressions.
- I want to add strings to line filters without using theApp.addToFilter.

sdottaka:
Sorry, I have reverted the addition of the HideFilterLines function.

The reason is as follows

- The line filter ignores differences in lines if the regular expression specified matches a line in both files. However, the HideFilterLines function ignores differences when only one of the files matches, so the behavior is different.
- The HideFilterLines function only applies to file comparisons, but WinMerge can also do folder comparisons. The behavior of the line filter must be the same for folder and file comparisons.

PR #1495, opened 2022-09-23
Fix "Left to Right" and "Right to Left" copying in 2-way folder comparison.

tjmprm77:
In the current version, when doing a "Left to Right" or "Right to Left" copy in 2-way folder comparison, the DIFFCODE::SAME flag is set for the copied item and all its descendants.
For this reason, if we perform a "Left to Right" copy for a directory that contains items that exist only on the right side, the comparison results are displayed as identical even though there is actually a difference in the directory after copying.

LeftToRight


sdottaka:
Thank you!

PR #1499, opened 2022-09-27
Fixed English verbiage

Gitoffthelawn:
Although this PR won't hurt anything, I'm not sure if this is the correct place to make this fix for all future releases.  If it's not the correct place, where would that be?


sdottaka:
Applied that fix to https://github.com/WinMerge/winmerge/releases/tag/v2.16.23

Gitoffthelawn:
You're welcome. As always, thank ***you!***

PR #1511, opened 2022-10-09
Update Brazilian.po

Marcello-mco:
Select Files or Folders Window: Make it possible to specify the Prediffer plugin


sdottaka:
Thank you!

PR #1513, opened 2022-10-10
Improved translation regarding CPU cores

Gitoffthelawn:


Gitoffthelawn:
Oooh... I didn't know that.  Given that info, the current string is more misleading than I originally thought, but my change is just as misleading! ;)

Let me think if there is some way to communicate this effectively.  Honestly, I think the UI will need to be changed to no longer employ negative numbers for this field.

If the user wants to automatically use all the available cores, what do they currently enter in that field?

sdottaka:
> If the user wants to automatically use all the available cores, what do they currently enter in that field?

If you enter 0, all available CPU cores are used.

Gitoffthelawn:
Okay... we can somehow figure this out!

If there are four CPU cores, then
4 uses 4 CPU cores.
3 uses 3 CPU cores.
2 uses 2 CPU cores.
1 uses 1 CPU core.
0 uses 4 CPU cores.
-1 uses 3 CPU cores.
-2 uses 2 CPU cores.
-3 uses 1 CPU core.
-4 is invalid.

Is this correct?

If so, do you think changing the UI to something like: "Number of CPU cores to reserve for non-WinMerge processes" will be more clear?

If so, the allowed values would then only be positive integers from 0 to (num_cores - 1).  No more negative core values!

sdottaka:
In commit 9c97fc5, I've changed "Number of compare threads (...)" to "Number of CPU cores to use".
To prevent negative values from appearing in the "Number of CPU cores to use" textbox, if the value stored in the registry is negative, the number of CPU cores will be added to the value displayed. 

image

Gitoffthelawn:
This works for me, but is it really what we want?  Besides manually editing the registry, the user will no longer be able to specify "use all available CPU cores except 2 of them", etc.

sdottaka:
To me, I find that input method confusing.
If there are 8 CPU cores, simply type in 6.
Note that if you enter -2, the registry will record -2 and the text box will show 6.

Gitoffthelawn:
I agree 100%.   The existing (before the pull request you made) is very confusing (hence the pull request I made).  With your pull request, we lose the functionality of allowing the user, via the WinMerge UI, to specify "use all cores except for 2 of them" (and other similar options).  That's a minor loss compared to the improved clarity of the UI, IMO.  I just want to make sure you're okay with that loss too.

sdottaka:
In fact, a user could have done the same thing as before by entering a negative value, but I think my change would have made the user unaware of that.
However, I don't think that would be a problem, so I decided it was OK.

Gitoffthelawn:
Sounds good to me too.

I am closing this issue now, as your PR will handle this.  If you think this issue needs to open for any reason, feel free to re-open it.

PR #1523, opened 2022-10-15
Update Brazilian.po

Marcello-mco:
Adding translation for Options dialog: Change "Number of compare threads (...)" to "Number o…


sdottaka:
Thank you!

PR #1535, opened 2022-10-23
Fix an issue where the message displayed when two files are identical in a 3-way folder comparison is not translated.

tjmprm77:
Fix an issue where the following message displayed when two files are identical in a 3-way folder comparison is not translated.
- (Middle and right are identical)
- (Left and right are identical)
- (Left and middle are identical)

The strings defined in Merge.rc have leading spaces, but the strings in DirViewColItems.cpp have no leading spaces and are not translated.
This PR fixes this issue by adding a leading space to the string in DirViewColItems.cpp.


sdottaka:
Thank you!

PR #1545, opened 2022-10-29
Changes the display processing of the "Comparison result" column for a 3-way folder comparison.

tjmprm77:
Add processing to indicate that two files are identical in the "Comparison result" column when one file does not exist and the other two files are identical in a 3-way folder comparison. 
compare_flags_3way


sdottaka:
Thank you!

PR #1559, opened 2022-11-06
Update Slovenian.po

JadranR:
Best Regards,
Jadran


JadranR:
Updated Slovenian translation, Today.
Best Regards,
Jadran

PR #1565, opened 2022-11-11
Add html5 keywords to crystaledit

stonee-K:
* Add html5 keywords (elements, attributes) to crystaledit
 * Obsolete and deprecated keywords in the current list was not removed.

reference:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes


sdottaka:
Thank you!

PR #1568, opened 2022-11-13
Fix the problem that WinMerge crashes when pressing the "OK" button in the "Display Columns" dialog in the debug version.

tjmprm77:
WinMerge crashes when the following steps are performed in the debug version.

Step 1. Open the "Display Columns" dialog and check only "Filename" and press the "OK" button.
Step 2. Reopen the "Display Columns" dialog and uncheck all and press the "OK" button.
Step 3. Open the "Display Columns" dialog again and press the "OK" button.

The cause is that the settings (OPT_DIRVIEW_COLUMN_ORDERS or OPT_DIRVIEW3_COLUMN_ORDERS) and the columns in the folder compare window are inconsistent after performing Step 2.
Settings: Revert to their default values
Columns in folder compare window: As they were before the Display Columns dialog was opened.

Due to this inconsistency, when the "Display Columns" dialog is opened in Step 3, the columns that are not displayed in the folder compare window but retained as settings are not registered in the "Display Columns" dialog.
This results in an access violation and crashes Winmerge.

This PR disables the "OK" button in the "Display Columns" dialog when no items are checked.
Also adds some error handling and assertions.

### DIFF
--- a/Src/DirColsDlg.cpp
+++ b/Src/DirColsDlg.cpp
@@ -250,15 +250,20 @@ void CDirColsDlg::OnOK()
 {
 	SanitizeOrder();
 
+	size_t colssize = m_cols.size();
 	for (int i = 0; i < m_listColumns.GetItemCount(); i++)
 	{
 		bool checked = !!m_listColumns.GetCheck(i);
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you!

PR #1572, opened 2022-11-16
Add css keywords to crystaledit

stonee-K:
* User 1 text color --> css1, css2.1, css3 keywords (standard)
 * User 2 text color --> experimental or deprecated keywords
 * Remove unknown keywords (ascent, bbox, ...)

reference:
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference


sdottaka:
Thank you!

PR #1574, opened 2022-11-16
Preliminary exit (performance optimization) for Scrollbars calculation

hordi:
Instead of loop all available lines to decide do we need to enable/disable the scrollbar - just enough to break when more than screen-lines analyzed.
Important for big files comparison.


sdottaka:
Thank you!

PR #1576, opened 2022-11-17
Fields rearranged for size optimization

hordi:
refactoring


sdottaka:
Thank you for the PR.

This PR had mixed SPACE and TAB in the indentation and the stringdiffs*. * files were incorrectly located.
I will correct this later.

PR #1577, opened 2022-11-17
refactoring Diff3.h

hordi:
refactor (::at to ::operator[]) replacement, light cleanup


sdottaka:
Thank you for your PR.

Argument diff3 is not supposed to be appended, so it's okay to assume size 0, and variable diff3count is unnecessary.
I will delete that variable later.

PR #1579, opened 2022-11-18
DiffFileData optimization for same-files

hordi:
DiffFileData: if 2 files are the same - use stat-information from first one
DiffItem::existAll, CompareStats.cpp - code lightly improved


sdottaka:
Thank you!

PR #1580, opened 2022-11-19
Fixed potentially wrong SubEditList ReadOnly attribute storage; refactoring

hordi:
Originally SubEditList uses std::set<bool> m_readOnlyColumns storage for read-only columns. Expectedly int-type should be used instead of bool-type if we want to make any particular (not only 0-index) column work in read-only mode or it doesn't have any sense to use std::set at all. Original code doesn't produce any issue because that function used in one place (I hope) with index 0 only, respectively 0-column only read-only and any other does not. Code refactored to eliminate usage of 2 redundant std::set storages for 2 different attributes and used one only std::vector to store and read_only+boolean_value type and EditStyle.

### DIFF
--- a/Src/SubeditList.cpp
+++ b/Src/SubeditList.cpp
@@ -27,6 +27,19 @@ CSubeditList::~CSubeditList()
 {
 }
 
+void CSubeditList::SetColumnAttribute(int nCol, int limit, int attribute)
+{
+	if (!IsValidCol(nCol))
+		return;
+
+	if (m_columsAttributes.size() <= static_cast<size_t>(nCol))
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you for the PR.

I think the name of the variable m_columsAttributes is a typo, so later I will change it to m_columnsAttributes.

hordi:
You right, thank you!

19 November 2022, 19:19:54, by "Takashi Sawanaka" ***@***.***>:


Thank you for the PR.
I think the name of the variable m_columsAttributes is a typo, so later I will change it to m_columnsAttributes.

PR #1586, opened 2022-11-24
unicoder.cpp light performance improvements

hordi:


sdottaka:
Thank you!

Src/Common/unicoder.cpp
```diff
@@ -792,10 +789,10 @@ void buffer::resize(size_t newSize)
 {
 	if (capacity < newSize)
 	{
-		capacity = newSize;
 		unsigned char *tmp = static_cast<unsigned char *>(realloc(ptr, capacity));
```
tjmprm77, line 792: I'm sorry if my understanding is wrong, but I think this process needs to be modified as follows.
```c
unsigned char *tmp = static_cast<unsigned char *>(realloc(ptr, newSize));
```

sdottaka: good catch. thank you!

Issue #1583, opened 2022-11-24
Syntax highlighting for SQL leaves many things out.


mdell-seradex:
I noticed when attempting to apply syntax highlighting (i.e. font formatting) for SQL files that many things were left unformatted.
Examples: many keywords such as "DECLARE", "CURSOR", "WHILE", "BEGIN", variable names, data types, the alias qualifier keyword "AS".
I was hoping to alter a configuration file to get it to recognize those, but I could not find one.

Is there a way we can modify this?  Alternatively, can this be updated to support more keywords, etc?

Thank you

mdell-seradex:
Cool.  Thanks for the update.
It is most appreciated.
I wasn't sure if I was missing something or not.
I notice that "DECLARE", "CURSOR", "FETCH", and "NEXT" weren't included.  Can they be added?
I get build a list of additional keywords, and possibly data types, if you think that would be useful.

sdottaka:
Pull request is welcome

sdottaka:
"DECLARE", "CURSOR", and "FETCH" were added by PR #1591.

PR #1591, opened 2022-11-29
Fix issue #1583 Syntax highlighting for SQL leaves many things out.

stonee-K:
* Add SQL:2016 keywords to crystaledit
* This keyword list does not contain non-standard keywords (GO, DUAL...)

reference:
https://en.wikipedia.org/wiki/List_of_SQL_reserved_words


sdottaka:
Thank you!

PR #1614, opened 2022-12-17
unify the translations of Filter

yingang:
also updated the left untranslated items after syncing with English.pot


sdottaka:
Thank you!

PR #1617, opened 2022-12-18
Update Brazilian.po

Marcello-mco:
Translate "Copy All Displayed Columns" to the context menu of the folder compare view.


sdottaka:
Thank you!

PR #1642, opened 2023-01-01
Merge-R2_8-gh

mrizwanazeez:
Comp-Updates.


sdottaka:
What does this PR fix?
In any case, the R2_8-gh branch is a very old version of WinMerge, so merging into this branch is not acceptable.

PR #1645, opened 2023-01-03
Modify the "Options (Compare > Folder)" dialog.

tjmprm77:
Disable "Include unique subfolders contents" when "Include Subfolders" is Off, because "Include unique subfolders contents" is only used when "Include Subfolders" is on.
Also, Swap the order of "Include unique subfolders contents" and "Include Subfolders".


sdottaka:
Thank you!

PR #1649, opened 2023-01-07
Add processing to indicate that two directories are identical in the "Comparison result" column when they are identical in a 3-way folder comparison. 

tjmprm77:
- Set the DIFFCODE::DIFF flag to directory items that don't have one or two directories when "Include unique subfolders contents" is on.
  (This process is also performed for 2-way folder comparisons.)
- Set the 3-way compare flag for directory items.
dir_3way_compare_flag


sdottaka:
Thank you!

PR #1667, opened 2023-01-14
sync zh-cn translation from English.pot

yingang:
only the last commit contains the actual PR content, didn't figure out why all the history commits will be attached :(


sdottaka:
Thank you!

PR #1683, opened 2023-01-22
Show/hide directories in 3-way comparison

tjmprm77:
Show/hide the directory with the following menu items in the folder compare window.
- "View" > "3-way Compare" > "Show Left Only Different Items"
- "View" > "3-way Compare" > "Show Middle Only Different Items"
- "View" > "3-way Compare" > "Show Right Only Different Items"


sdottaka:
Thank you!

PR #1711, opened 2023-02-10
 Installer: Update Brazilian.isl

Marcellomco:
Added translation for Corsican language option and corrected "DeletePreviousStartMenu" entry, there were two percent signs missing.


sdottaka:
Thank you!

PR #1712, opened 2023-02-12
Fix an issue where some messages are not translated.

tjmprm77:
Fix an issue where the following messages are not translated.
- "Cannot find file filter template file!\n\nPlease copy file %1 to WinMerge/Filters folder:\n%2."
- "Left side file version, only for some file types."
- "Right side file version, only for some file types."
- "Middle side file version, only for some file types."


sdottaka:
Thank you!

PR #1713, opened 2023-02-15
Update Brazilian.isl

Marcellomco:
Added translation for "Installer: Rename the ShellExtension*.dll file names before installin…"


sdottaka:
Thank you!

PR #1751, opened 2023-03-12
Fix invalid msgstr in (Russian.po)

giga-bytes-dev:
Fix invalid msgstr for Don't ask question (Russian.po)


sdottaka:
Thank you!

PR #1779, opened 2023-03-26
Modify the "Go to" dialog.

tjmprm77:
Disable the "Middle" radio button when doing 2-way file comparisons.


sdottaka:
Thank you!

PR #1806, opened 2023-04-16
Update Brazilian.po

Marcellomco:
Translation for 'Make the plugin settings dialog translatable and enable saving settin...'


sdottaka:
Thank you!

PR #1826, opened 2023-04-30
Fix input range check processing in "Go to" dialog.

tjmprm77:
Fix the following issues:
- When "Go to what" is "Line", specifying a line greater than the number of lines in the selected pane when opening the dialog navigates to the wrong line.
  (Navigate to a line equal to the number of lines in the pane selected when the dialog was opened.)
- When "Go to what" is "Difference" in debug build, if a number larger than the number of comparison differences is specified, a debug error occurs in the CMergeEditView::SelectDiff() check process.


sdottaka:
Thank you!

PR #1827, opened 2023-05-01
Confirm copy all in file merge

wilfz:
WinMerge is my new favorite diff and merge tool. Thank you. 
But unfortunately it easily happens to copy all diffs to the other file instead of one specific difference without noticing it.
To avoid this I added a new "Are you sure you want top copy ALL diffs to other file?" dialog. The dialog can be turned off like the other dialogs.
For localization I also added the translation to my native language german and the Spanish.po file.


sdottaka:
Thank you!

sdottaka:
In commit cccf9c8 I changed "Are you sure you want top copy ALL diffs to the other file?" to "Are you sure you want to copy all differences to the other file?". Please let me know if you see any problems with this change.

PR #1873, opened 2023-05-27
Translate some plugin error messages

tjmprm77:
Make the following plugin error messages translatable.
- "Plugin not found or invalid: %1"
- "'%s' is not PACK_UNPACK plugin"
- "'%s' is not PREDIFF plugin"
- "An error occurred while prediffing the file '%s' with the plugin '%s'. The prediffing is not applied any more."


sdottaka:
Thank you!

PR #1876, opened 2023-05-28
[XML] Update Brazilian.po

Marcellomco:
Added translation for #1873: Make the following plugin error messages translatable.


sdottaka:
Thank you!

PR #1896, opened 2023-06-11
Modify the "Go to" dialog.

tjmprm77:
The following modifications are made to the "Go to" dialog.
- Show the acceptable range for the selected "File" and "Go to what".
- Disable the "Go to" button if the input form does not contain a value within the acceptable range.
- Disable the "Difference" radio button when the files have no differences (i.e. identical files).


sdottaka:
Thank you!

PR #1938, opened 2023-07-17
Update Python syntax highlighting keyword list.

tjmprm77:
Added some keywords defined in version 3.11.4.
- https://docs.python.org/3/reference/index.html
- https://docs.python.org/3/library/index.html


sdottaka:
Thank you!

PR #1945, opened 2023-07-23
Allow changing the number of CPU cores to use while doing folder comp…

sdottaka:
…arisons

This PR adds a "Number of CPU cores to use" combobox to the folder comparison progress window, allowing you to change it during the comparison, as shown below.

image


sdottaka:
@vivadavid The explanation here is insufficient, but if a comparison method other than "Full Contents" or "Quick Contents" is selected, using multiple CPU cores will actually slow down the process. Therefore, the combo box is restricted to allow only the selection of "1".

Furthermore, when specifying "Full Contents" or "Quick Contents" as the comparison method, values greater than those specified in the Options window cannot be used.

vivadavid:
I didn't know that: thank you for the information, @sdottaka !

PR #1948, opened 2023-07-26
Update Brazilian.po

Marcellomco:
Translation for "Allow changing the number of CPU cores to use while doing folder comp…"


sdottaka:
Thank you!

PR #1969, opened 2023-08-05
Update Brazilian.po

Marcellomco:
Added translation for "Add Expand Different Subfolders menu item (#1964) "


sdottaka:
Thank you!

PR #1998, opened 2023-08-27
Update Rust syntax highlighting keyword list.

tjmprm77:
Added some keywords defined in version 1.72.0.
- https://doc.rust-lang.org/book/appendix-01-keywords.html
- https://doc.rust-lang.org/std/index.html


sdottaka:
Thank you!

PR #2000, opened 2023-08-28
Optimize inserts in std containers using reserve

GermanAizek:
@sdottaka hi.
I think you will need such optimizations that affect Debug and Release build.


sdottaka:
Thank you! 
I'm concerned about the reserve() argument in the CreateVirtualLineToRealLineMap3way() function in the MergeDocDiffSync.cpp file, which I may fix later.

PR #2063, opened 2023-10-07
Update English.pot for correct translation in others languages

Need74:
Hi
These two small modifications make it possible to correctly translate WinMerge into another language because otherwise these two occurrences desperately remain in the original language on the interface despite the contry.po file corresponding to its user language.
 
I also think that the expression "E&xpand All Subfolders" like "&Collapse All Subfolders" are not explicit in the winmerge.exe binary and that it would have been more judicious to name these variables "E&xpend TreeView" and &Collapse TreeView ".

I also noted that the line "Diff &Algorithme (Experimental)" changed to "Diff &Algorithme" on version 2.16.32 and that as a result the translations of the text are no longer effective on this occurrence. We should ask the developers if it is a desire to remove the notion (Experimental) or if it is a simple typing error in english.pot

I am aware that what I express lacks clarity and technical rigor but I am neither a developer nor a computer scientist but just a humble translator.

Sincerely

### DIFF
--- a/Translations/WinMerge/English.pot
+++ b/Translations/WinMerge/English.pot
@@ -469,7 +469,7 @@ msgstr ""
 msgid "Tree &Mode"
 msgstr ""
 
-msgid "E&xpand Subfolders"
+msgid "E&xpand All Subfolders"
 msgstr ""
 
 msgid "&All Subfolders"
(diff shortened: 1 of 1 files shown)


sdottaka:
Please note that the files in the current repository are in development for version 2.16.34, not version 2.16.32.

The "Expand All Subfolders" menu item was changed in 2.16.33 as follows.
image

Also, in 2.16.33, "Diff algorithm (Experimental)" was changed to "Diff algorithm". This is because in 2.16.32, it was only applied to file comparisons, but in 2.16.33, it is now applied to folder comparisons as well.

Need74:
Thank you for the comment

I understood and I will be careful now

PR #2068, opened 2023-10-15
Modify the "Substitution Filters" dialog.

tjmprm77:
- Disable the "Remove" button when no item is selected.
- Disable the "Clear" button when no items are registered in the list.


sdottaka:
Thank you!

PR #2079, opened 2023-10-29
Update English.pot

Need74:
Hi
to make possible the translation into French (or another language) of the "print preview" window, I tried without success to add the few lines of the missing words in my french.po file but this had no effect. This did not surprise me too much since these words are absent from the English base english.pot but it may not be enough to declare this translation if the binary is not up to date for this. It is possible that my proposal is wrong and in this case I apologize but it would be nice to be able to translate the print preview page into all languages. Kind regards, Denis

PS I haven't found anywhere a possible configuration for my WinMerge to detect updates and I am therefore obliged to visit GitHub from time to time to check for new features.

### DIFF
--- a/Translations/WinMerge/English.pot
+++ b/Translations/WinMerge/English.pot
@@ -4088,3 +4088,20 @@ msgstr ""
 msgid "Do not specify the '-p0' command line option for the patch file which includes absolute paths"
 msgstr ""
 
+msgid "&Print"
+msgstr ""
+
+msgid "&Next Page"
+msgstr ""
+
+msgid "Pre&v Page"
+msgstr ""
+
+msgid "Zoom &in"
+msgstr ""
+
+msgid "Zoom &out"
+msgstr ""
+
+msgid "&Close"
+msgstr ""


sdottaka:
Print Preview is displayed by the MFC library, so it seems that some effort is needed to make it translatable.
Also, English.pot is a file that is automatically generated from Merge.rc, so it is not appropriate to modify this file.

Need74:
Hi
thank you for your response and I note english.pot is not an editable file.
However, I have found errors when translating the software into French and sometimes my translation problem is linked to the fact that the english.pot model does not allow the function of a command or command to be intelligently translated. 'a control button.
On this subject I discovered an interpretation error between Left, Right, Midle, and on the left, on the right, to ou from the middle, which led to keyboard shortcuts being entered in the menus or tooltips which are not correct because they depend on the panel where it is located. the insertion point when using in 3-panel mode.
Thank you for telling me how I can make my observation and also my proposal for correction, knowing that this is only a problem with the interface but not functionality.
Sincerely
Dennis

sdottaka:
First, I cannot apply this Pull Request, so I will temporarily close it.
Regarding the issue of the untranslated "Print Preview" buttons, I have created a new issue #2083.

 >  On this subject, I discovered an interpretation error between "Left," "Right," "Middle," and "on the left," "on the right," "from or towards the middle," which led to keyboard shortcuts being included in the menus or tooltips that are not correct. In fact, these shortcuts depend on the panel where the insertion point is located when used in the 3-panel mode.

Regarding this matter, it is a separate issue from Print Preview, so could you create a new issue and provide specific details about your expectations and the actual results?
Additionally, it might be challenging to change menus and tooltips based on the selected pane.

sdottaka:
Commits a0abf31 and d021044 made print preview buttons translatable.

PR #2085, opened 2023-11-03
Galician and Spanish translations update

qosobrin:
Here you have my updates to the Galician and Spanish translations


sdottaka:
Thank you!

PR #2086, opened 2023-11-05
fix bug of Registry path

Yutaka-Sawada:
Per-user installer has a bug, which makes registry keys in wrong position. 

From Inno Setup Help;
> Using HKCR is not recommended, use HKA with the Subkey parameter set to "Software\Classes" instead.

Setting `HKCU` only is different from `HKCR`. 
Adding `Software\Classes` before creating items is good.


sdottaka:
Thank you!

PR #2088, opened 2023-11-06
Update Brazilian.po

Marcellomco:
Added translation for " Make the Preview bar buttons translatable ("Two Page" and "One Page" are not yet translatable...)"


sdottaka:
Thank you!

PR #2089, opened 2023-11-06
Romanian translations update

drdcd:


drdcd:
The Romanian translations had not been updated for the past three years or so and were utterly incomplete. For InnoSetup I've additionally replaced the ANSI with a Unicode file.

sdottaka:
Thank you!

PR #2118, opened 2023-11-23
Modify the "File Filters" dialog.

tjmprm77:
Disable the "Test", "Edit" and "Remove" buttons when no filter is selected in addition to when "None" filter is selected.


sdottaka:
Thank you!

PR #2136, opened 2023-12-14
Added new C# keywords

myss:
Keywords are from https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/


sdottaka:
Thank you!

PR #2151, opened 2023-12-28
Update Brazilian.po

Marcellomco:
Added translation for "Webpage Compare: Display "Comparing..." in the status bar during comp…"


sdottaka:
Thank you!

PR #2154, opened 2023-12-29
Modify the "Display Columns" dialog.

tjmprm77:
- Clear the column description and disable "Move Up" and "Move Down" buttons when no item is selected.
- Select the first item after clicking the "Defaults" button.


sdottaka:
Thank you!

PR #2166, opened 2024-01-04
Update C++ syntax highlighting keyword list.

tjmprm77:
- Keywords are from https://en.cppreference.com/w/cpp/keyword
- Remove some duplicate keywords from the list.


sdottaka:
Thank you!

PR #2211, opened 2024-02-08
Update Russian.po

Timon34:
Uniformity of translation of the same phrase in different sources (InnoSetup and ShellExtension)


sdottaka:
Thank you!

PR #2215, opened 2024-02-12
Update Java syntax highlighting keyword list.

tjmprm77:
Add some keywords defined in Java SE 21 Edition.
- Keywords are from https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-3.9


sdottaka:
Thank you!

PR #2230, opened 2024-02-29
Replace "Match similar lines" with "Align similar lines"

vlakoff:
This PR is a follow-up to #1200.

* One commit is for the documentation, so that it matches the setting name that is displayed in the program.
* The other commit is for the program source, better make things consistent.

### DIFF
--- a/Docs/Manual/English/Compare_files.xml
+++ b/Docs/Manual/English/Compare_files.xml
@@ -289,14 +289,14 @@
               </indexterm></term>
 
             <listitem>
-              <para>WinMerge tries to match similar lines within difference
-              blocks when you enable <option>Match similar lines</option> in the
+              <para>WinMerge tries to align similar lines within difference
+              blocks when you enable <option>Align similar lines</option> in the
               Options dialog Compare page. What does <quote>similar</quote>
               mean? Generally, this feature works best in simple cases, for
               example where lines that have strong similarities. When detected,
               similar lines are adjusted to align in the File pane. The next
               figure shows comparisons of the same two files, before and after
-              enabling the <option>Match similar lines</option> option:</para>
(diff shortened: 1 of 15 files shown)


sdottaka:
Thank you for the PR!
I have merged this PR.
However, since the compatibility of the configuration information will be lost, I would like to restore part of Line 212 of OptionsDef.h to the following.

~~~
inline const String OPT_CMP_ALIGN_SIMILAR_LINES {_T("Settings/AlignSimilarLines"s)};
~~~
↓
~~~
inline const String OPT_CMP_ALIGN_SIMILAR_LINES {_T("Settings/MatchSimilarLines"s)};
~~~

vlakoff:
Could you please elaborate on what could have been the issue? From what I understood, it is a constant so the actual value it contains isn't important, what we use is the symbol.

By the way, there is a remaining entry that you may want to replace as well: the `<title>類似行をマッチさせる<indexterm>`, [in Docs/Manual/Japanese/Configuration.xml](https://github.com/WinMerge/winmerge/blob/c15957ba0a92fad6ac072c5dfef56e4cfdceaccf/Docs/Manual/Japanese/Configuration.xml#L582).

sdottaka:
The string above is the key name used to save or load the “Align similar lines” checkbox value to the registry or INI file.
If this is changed, the "Align similar lines" checkbox will revert to an unchecked state in the new WinMerge version, even if the "Align similar lines" checkbox was checked in an earlier WinMerge version.

vlakoff:
Understood, thanks for the clarification.

Agree that keeping the current configuration item name is the best to do. It should not be very bad, it seems there are other configuration items in the same case, having functionnality that has slightly diverged over time from the unchanging configuration item name.

PR #2234, opened 2024-03-04
Update Brazilian.po

Marcellomco:
Added translation for "Make it possible to select the behavior when copying to another pane with 'Copy to Right' etc. while text is selected. (#2224)"


sdottaka:
Thank you!

PR #2243, opened 2024-03-08
Update Brazilian.po

Marcellomco:
Added translation for "Make it possible to transform the text copied with 'Copy Right/Left' using a plugin. (#2238)"


sdottaka:
Thank you!

PR #2247, opened 2024-03-08
Move custom messages in a separate iss file

bovirus:
@sdottaka 

I moved Inno Setup custom messuges (now english/italian - equal for all .iss installer files) in a separate file

**WinMergeCustomMessages.iss**

The it'eseasier for anyone add new language strings without touch main installer script.

And the changes are required in only one file.
befroe these changes you have to apply the custom emssage strings for each .iss installer script.

### DIFF
--- a/Installer/InnoSetup/WinMergeARM64.is6.iss
+++ b/Installer/InnoSetup/WinMergeARM64.is6.iss
@@ -53,6 +53,7 @@
 #define AppVersion GetFileVersion(SourcePath + "\..\..\Build\" + ARCH + "\Release\WinMergeU.exe")
 #define ShellExtensionVersion GetFileVersion(SourcePath + "..\..\Build\ShellExtension\" + ShellExtension64bit)
 #define WinMergeContextMenuVersion GetFileVersion(SourcePath + "..\..\Build\ShellExtension\" + ARCH + "\WinMergeContextMenu.dll")
+#include WinMergeCustomMessages.iss
 
 [Setup]
 AppName=WinMerge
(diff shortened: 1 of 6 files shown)


sdottaka:
Thank you!

bovirus:
@sdottaka 

Please add in this file

https://github.com/WinMerge/winmerge/blob/master/Translations/README.md

the info about how to translate custom messages for installer.

The current note about Installer translation can create misunderstanding.
The installer are already multilanguage (use of .isl files) and require only activities by translator about custoim messages.

Now the custom emssges are in a separate file (out of main installer script) and translator can add other languages (now are only english anbd italian).

Thanks.

sdottaka:
Regarding your PR, I got an error when compiling with InnoSetup. It seems you forgot to add double quotes.
I found that the messages you moved to WinMergeCustomMessages.iss can also be moved to the .islu file, so
1a9f0fa moved the messages to those files.
I also deleted messages that I felt didn't require customization.

sdottaka:
>For me is not a good idea to modify original .isl file (coming form Inno Setup standard installation) adding custom messages that are specific for the project.

I may have misunderstood your response, but Translations/WinMerge/*.isl stores messages used only by the WinMerge installer, not files that came from the Inno Setup standard installation. 

> My idea is that .isl file should remain untouched and Custom Messages strings should be in a separate file (single file for all languages).

I feel it is better for translators to work with just one file for the installer rather than two.
Also, WinMerge supports 37 languages, but if 37 languages are stored in one file, translators will not be able to edit them at the same time.

PR #2249, opened 2024-03-09
Update PO header for Italian files

bovirus:
@sdottaka 

I updated the header of Italian language files to get the info about relase date for statistics.


sdottaka:
Thank you!

PR #2265, opened 2024-03-20
Update PHP syntax highlighting keyword list.

tjmprm77:
Add some keywords defined in PHP 8.3.
- Keywords are from https://www.php.net/manual/en/reserved.php


sdottaka:
Thank you!

PR #2288, opened 2024-04-21
Update Corsican translation for 2.16.40

Patriccollu:
Hello,

This is an update of **Corsican** (co) localization to take the following commit into account:

 * https://github.com/WinMerge/winmerge/commit/768a6fa7a2d1f12554c0716ce74059b54ebbf2bf Allow plugin pipeline aliases or simple plugins to be registered in the GUI (#2257)

Best regards,
Patriccollu.


sdottaka:
Thank you!

PR #2295, opened 2024-04-23
Adjust position of "Defaults" button in "Options > Compare > Table"

vlakoff:
This way, the "Defaults" button has exactly the same position on all subsections (General, Folder, Table, etc.) of the Compare section, in the program settings.

That position offset had been introduced in [this change](https://github.com/WinMerge/winmerge/commit/57696bba6d21c53b82eb730c613b6a741d7feb0e#diff-d5fc4f096c1ae14bcea4ac7b6bf65cf6b8a92a99d231fff653d0e7fc84395ea2L2216). Apparently to horizontally align with the new "quote character" text input. Though, this horizontal alignment doesn't bring much benefits, whereas the UI is much better if the button has the same position on all pages.


sdottaka:
Thank you!

PR #2298, opened 2024-04-27
Added support for F# language syntax highlight

Thorium:
Using the existing CSharp syntax highlight as basis, added support for FSharp language syntax highlight.


sdottaka:
Thank you!

sdottaka:
It seems that block comments `(* *)` are not supported, so I would appreciate it if you could fix it if you have time.

Thorium:
Well spotted! I can look into these when I'll have a bit of spare time.

PR #2301, opened 2024-04-28
Improved F# parsing, support for multi-line comments

Thorium:
Effort to improve current initial F# parsing, support for multi-line comments. 
ParseLineCJava was used as base for ParseLineFSharp.

### DIFF
--- a/Externals/crystaledit/editlib/parsers/fsharp.cpp
+++ b/Externals/crystaledit/editlib/parsers/fsharp.cpp
@@ -150,6 +150,36 @@ static const tchar_t * s_apszFsKeywordList[] =
     _T ("yield!"),
   };
 
+  static const tchar_t* s_apszUser1KeywordList[] =
+  {
+    _T("Some"),
+    _T("None"),
+    _T("ValueSome"),
+    _T("ValueNone"),
+    _T("__SOURCE_DIRECTORY__"),
+    _T("DateTime"),
+    _T("String"),
+    _T("Decimal"),
+    _T("Char"),
+    _T("Int16"),
+    _T("Int32"),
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you for the PR. However, block comments were not colored correctly, so I fixed it in commit 131a7e2.
Also, as shown in the image below, it seems that there are improvements to be made, such as not supporting F#'s raw string literal, so I would appreciate it if you could fix it when you have time.

image

Thorium:
I'm often using WinMerge to compare F# code files, IMO WinMerge does the 2-way comparison better than e.g. KDiff. I think this kind of basic F# syntax highlighting is very nice addition already.

But I think it's not worth of implementing full F# parser independently: Lot of F# development relies on tooltips (check e.g. [here](https://fssnip.net/ml/title/Small-quizgame)), the compiler telling you the types. That is typically done in editors by referencing F# compiler service which parses the F# for you. As general IDE, I don't think WinMerge will want this dependency, people will probably not use WinMerge as primary F# editor anyway.

Rather than creating all parsers separately, in general one possible better long-term future addition for WinMerge 
could be supporting [tree-sitter](https://tree-sitter.github.io/tree-sitter/). That would make it easier to parse any programming language and having big community effort behind (and there is also [tree-sitter for F#](https://github.com/ionide/tree-sitter-fsharp)).

Thorium:
I tried to add raw string literal support https://github.com/Thorium/winmerge/commit/1cf5458d6fe19c9f4cd81408a8b50b0fec9eae59 but it's not working... I have to say my C++ is a little bit rotten as I've not used it in last 25 years.

PR #2303, opened 2024-04-30
Multi line comment fix

Thorium:
Work continued from https://github.com/WinMerge/winmerge/pull/2301
It does work already better than what is in master now:

image

... but it's not exactly perfect, as you can see the first line of multi-line is gray (here the last quote of the multi-line).


sdottaka:
Thank you!

PR #2308, opened 2024-05-01
Update Brazilian.po

Marcellomco:
Added translation for "Add a text conversion plugin using OpenAI API (#2258)"


sdottaka:
Thank you!

PR #2310, opened 2024-05-02
re-update zh-cn translation

yingang:
ps. don't know why the previous checkin has been rolled back


sdottaka:
very sorry. It seems I made a mistake in the git operation and reverted it with commit 83e5afe...

yingang:
get it and thanks for your prompt repsonse!

PR #2312, opened 2024-05-04
Typescript files to use JavaScript highlighting as default

Thorium:
Typescript files to use JavaScript highlighting as default.


sdottaka:
Thank you!

PR #2349, opened 2024-06-10
Uliahn patch translations

uliahn:
Add OpenAI related translations. 
Add some missing translations. 
Not complete yet. 

Todo: Check accelerator hotkeys "&". Must occur only once on one page or dialogue.


sdottaka:
Thank you!

PR #2362, opened 2024-06-23
Add a "Defaults" button to the "Options (Archive Support)" dialog.

tjmprm77:
Add a "Defaults" button to the "Options (Archive Support)" dialog.
dialog


sdottaka:
Thank you!

PR #2364, opened 2024-06-27
Enable Rounded Corners for WinMerge Menu on Windows 11

sdottaka:
This PR fixes the issue where WinMerge menus do not have rounded corners on Windows 11 by removing the owner draw.
However, there is still an issue with radio items not being drawn properly, which has not been resolved.

WinMerge version 2.16.40
image

WinMerge version 2.16.42 (this PR)
image

### DIFF
--- a/Src/Common/BCMenu.cpp
+++ b/Src/Common/BCMenu.cpp
@@ -51,11 +51,7 @@ int BCMenu::m_checkBgWidth = 0;
 int BCMenu::m_gutterWidth = 0;
 int BCMenu::m_arrowWidth = 0;
 HTHEME BCMenu::m_hTheme = nullptr;
-#ifdef _WIN64
-bool BCMenu::m_bEnableOwnerDraw = false;
-#else
 bool BCMenu::m_bEnableOwnerDraw = true;
-#endif
 
 static class GdiplusToken
 {
--- a/Src/Common/BCMenu.cpp
+++ b/Src/Common/BCMenu.cpp
@@ -51,6 +51,11 @@ int BCMenu::m_checkBgWidth = 0;
 int BCMenu::m_gutterWidth = 0;
 int BCMenu::m_arrowWidth = 0;
 HTHEME BCMenu::m_hTheme = nullptr;
+#ifdef _WIN64
+bool BCMenu::m_bEnableOwnerDraw = false;
+#else
+bool BCMenu::m_bEnableOwnerDraw = true;
+#endif
 
 static class GdiplusToken
 {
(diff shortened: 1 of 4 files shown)


MagicAndre1981:
Can you please add an option to disable this? I don't like rounded corners

sdottaka:
Could you create an issue about that? We don't have plans to make it switchable at the moment, but if many people dislike it, we may change it back.

By the way, can you articulate how you dislike rounded corners?

For example, in some applications, rounding the corners can cut off text, or in the case of images, the images can be cut off. Or it may cause unnecessary CPU load.

mdonatas-trafi:
Rounded corners in menus is OK, but tabs and text backgrounds (such as `Location pane`) look very weird.

I cast my vote to **revert this** (except menus).

_Looking at how to build my own executable with this reverted for the time being 😅_

sdottaka:
Thank you for your feedback.

I understand that you find the rounded corners in menus acceptable but feel that tabs and text backgrounds (such as the Location pane) look strange. While I don't intend to make major changes at this point, I'm curious about why you feel this way.

Do you think the issue is the rounded corners themselves, or could it be that the corner radius is too large or that the small spaces and shadows introduced with the rounded design look off?

Additionally, do you find the rounded corners on other platforms like GitHub or many other websites bothersome? If not, could it be more of a balance issue rather than the rounded corners themselves?

mdonatas-trafi:
@sdottaka first of all, thanks for all the work that you do, while I was bisecting to see where there changes affecting each "rounding" were I was surprised by how much of the code had to change for that.

The main reason for such a reaction was discontinuity of the UI. If you look at this, everything is cramped, toolbar buttons have no margins and probably only ~2px padding - and I like it! This means there is more space for the important stuff - the diff.
image

In the updated UI we loose 1 row of diff for no good reason (in my biased opinion).
image

`1` - this feels artificially shrunken.. as if "the container is clearly WAY too small". I understand that amount of text visible is the same on both cases but the previous version feels roomier.
`2` - This irks me because it represents visible rows and I see whole rows and the diffs view-port is not rounded and thus why should it's representation be? It also has discontinuity in that the rectangle representing full document length doesn't have rounded corners (please don't.. better revert the rounding of this particular rectangle)
`3` - Way too much attention grabbing. Again, the main thing why this program was opened is a diff. No need to bold it.

p.s. It was a bit sad that in the code changing `RR_RADIUS` didn't just change the radius of the corners but actually changed the dimensions of the rectangle itself.. this ruined my hopes of an easy fix to get the rectangles back :)

sdottaka:
Thanks for your feedback.

I'm also a little concerned about the narrower display area for the diffs. For this reason, I'd like to move the tabs to the title bar like other web browser applications.

1. I'm thinking of increasing the margins a bit.
2. I rounded the corners because I wanted it to look like it was in a spotlight. I probably won't round the white areas.
3. Some of the text was hard to see, probably because of the background color, so I tried making it bold and it looked better to me. The main focus is certainly the diff, but shouldn't you focus on what the diff is in the first place?

PR #2376, opened 2024-07-12
Add Support for Custom System Color Overrides

sdottaka:
This PR hooks the Windows system color retrieval function GetSysColor() to return customized colors. This allows you to change the color of toolbars, status bars, etc. as shown below. However, this only works in some areas and does not currently apply to dialog windows, scroll bars, etc.

image

You can customize the system colors in the Colors/System category of the Options window.

image


Gitoffthelawn:
Do you think it will be possible to override scrollbar colours at some point?

PR #2391, opened 2024-07-24
Update Brazilian.po

Marcellomco:
Added translation for "Image Compare: Make it possible to change the blink interval and over…"


sdottaka:
Thank you!

PR #2402, opened 2024-08-06
Drop unused `*POFiles.bat` files

gerundt:
The `InsertLineNumberInPOFiles.bat` and `RenewPOFiles.bat` files call non exsiting scripts, so I think they are unused we can remove them. 😉


sdottaka:
Ah, it seems I forgot to fix those batch files when I converted VBScript to JScript.
I'll apply this PR for now, but if it seems like it can be fixed, I might revert it.

sdottaka:
In commit 75d0d05, InsertLineNumberInPOFiles.bat and RenewPOFiles.bat were made to work.

PR #2429, opened 2024-09-16
Copy Selected Difference to Clipboard

lededev:
Add shortcut Ctrl+1 .. 3 Copy selected diffrence from Left/Mid/Right view to system clipboard.
image
image


lededev:
Thanks, I choose to use the Ctrl shortcut because the next operation is usually to paste Ctrl+V, so the finger moves a shorter distance.

PR #2431, opened 2024-09-16
Update Brazilian.po

Marcellomco:
Added translation for "Add Copy Selected Difference to Clipboard (#2429)"


sdottaka:
Thank you!

PR #2446, opened 2024-09-21
Update Italian language

bovirus:
@sdottaka 

Please merge.

Note: in .po file there are several "/r" code that are not standrd for .po files.
The standard is only "/n".
Please check and remove them if you can.


sdottaka:
Thank you!

This may need to be improved, but for now some messages need to be \r\n.
Please note that if you change \r\n to \n, the line breaks will not be correct as in the attached image.

image

PR #2454, opened 2024-09-29
Update Brazilian.po

Marcellomco:
Added translation for "Add support for placing the tab bar on the title bar!


sdottaka:
Thank you!

PR #2489, opened 2024-10-23
Easy move small main window when tab on titlebar

lededev:
When the titlebar is filled with tabs or the main window is too small, there's no space left for dragging and moving the window. This PR allows the rightmost third of the system menu icon area to be used for dragging the window.
image


sdottaka:
Thank you!

PR #2497, opened 2024-10-27
Foreground and background title bars are distinguishable

lededev:
Currently, with multiple WinMerge instances open, it’s difficult to identify the active window. This PR changes the title bar background color when the program loses focus, making it easier to spot the active window among multiple instances.
Before
屏幕截图 2024-10-27 214315
After
f20241027T215958
With colorful picture wallpaper, the distinction between the two backgrounds remains low. However, I believe color adjustments should not be decided by me, so I am only submitting the functional code.

### DIFF
--- a/Src/Common/MDITabBar.cpp
+++ b/Src/Common/MDITabBar.cpp
@@ -9,6 +9,7 @@
 #include "IMDITab.h"
 #include "cecolor.h"
 #include "RoundedRectWithShadow.h"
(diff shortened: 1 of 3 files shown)


sdottaka:
Thank you for the PR.
I would prefer not to include MainFrm.h in MDITabBar.cpp if possible, so I may change it later so that it does not include MainFrm.h.

PR #2505, opened 2024-10-28
Title bar background color and text color follow windows settings when not customized

lededev:
When not check on this option
image
Run WinMerge without command line params, focus and no focus, enable tabs on title bar, title bar bgcolor will follow windows settings COLOR_ACTIVECAPTION and COLOR_INACTIVECAPTION, title text wll follow windows settings COLOR_CAPTIONTEXT and COLOR_INACTIVECAPTIONTEXT
f20241028T230641
WinMerge with tabs focus and no focus
f20241028T230830

### DIFF
--- a/Src/Common/MDITabBar.cpp
+++ b/Src/Common/MDITabBar.cpp
@@ -10,6 +10,7 @@
 #include "cecolor.h"
 #include "RoundedRectWithShadow.h"
 #include <dwmapi.h>
+#include <RegKey.h>
 #pragma comment(lib, "dwmapi.lib")
 
 #ifdef _DEBUG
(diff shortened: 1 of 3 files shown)


sdottaka:
Thank you for your PR.

The background color of the title bar is a slightly bluish color of COLOR_3DFACE, and the text color is the color of COLOR_WINDOWTEXT, so I think you can customize the title bar color by changing these colors.

(However, COLOR_3DFACE is also used for the status bar, etc., so changing it will also affect the status bar, etc.)

However, since you submitted this PR, does that mean you want to change the title bar color separately from the status bar, etc.?

If so, you can generally set the title bar color to the accent color by turning on Show accent color on the title bars and window borders in Settings → Personalization → Colors, so it may be a good idea to load this setting and use the accent color in WinMerge as well.

lededev:
I just set WinMerge's title color the same as my windows system settings.
image

lededev:
> However, since you submitted this PR, does that mean you want to change the title bar color separately from the status bar, etc.?
> 
> If so, you can generally set the title bar color to the accent color by turning on Show accent color on the title bars and window borders in Settings → Personalization → Colors, so it may be a good idea to load this setting and use the accent color in WinMerge as well.

When enable tabs on titlebar, the title bar is draw by app not the OS, without this PR, title color will not follow system COLOR_ACTIVECAPTION settings, it always COLOR_3DFACE, becasue it is hard-coded in `COLORREF CMyTabCtrl::GetBackColor() const`. If disable tabs on titlebar, windows OS can draw the title bar using system settings color.

lededev:
Microsoft Edge's title bar follow windows settings, not using COLOR_3DFACE but COLOR_ACTIVECAPTION
image

lededev:
@sdottaka  I reversed the rule, in the case of non-customization, the color is displayed by Windows settings, and the customization remains the same as before.

lededev:
Since Tabs on title bar is a new feature, it's not perfect, and it breaks the habits of some old users, so much so that some users ask for it to be changed back to the original feature. My view is to try to improve it rather than go back. For large monitor users who are used to multi-instances rather than multi-tabs, quickly positioning to the currently active window from multiple windows is a basic need, not just making it look cool.

lededev:
All using windows settings
image

PR #2506, opened 2024-10-28
Add a "Defaults" button to the "Options (Colors/System)" dialog.

lededev:
image


sdottaka:
Thank you for the PR.

I thought that the Defaults button was unnecessary, because when you want to set the settings on each page in the Colors category to the default, you can simply reselect the color scheme in the Colors/Color Schemes category.

If you were to add a Defaults button, I think it would be better to set default values ​​according to the currently selected color scheme when you press the Defaults button. (This PR does not do this. For example, if you press the Defaults button while the Solarized Dark color scheme is selected, the colors of the status bar etc. will not be those of Solarized Dark.)

lededev:
If you change the color value of the COLOR_ACTIVECAPTION and save it, the change will not be overwritten by switching the color scheme, it will be saved in the "Settings/SysColorHookColors" in the registry, and the default value will be overridden every time it is loaded, which will be overridden by the void PropSysColors::ReadOptions() function. If you change all the color values in the drop-down box, it is very cumbersome to change back to the default values, so the "Defaults" button is necessary.

sdottaka:
I've merged it for now.
However, as I mentioned before, I want the default values ​​to reflect the current color scheme, so I plan to modify them later.

PR #2509, opened 2024-10-29
Main Menu toggle Show/Hide

lededev:
Now we can have the simplest user interface, suitable for keyboard-centric power users and small screen users, maximizing the content display area.
The hotkey to show and hide the main menu is Shift+F12. You can also right-click on the title bar or left-click on the system menu icon to select from the system menu. Even with the main menu hidden, the main menu also works normally. For example, pressing Alt+F will bring up the File menu, Alt+E will bring up the Edit menu, and Alt+V will bring up the View menu. So there are other ways to restore the main menu, such as pressing (Alt+Space, H) or (Alt+V, H) or (F10, V, H).

Here is the display effect after hiding the main menu, tab bar, toolbar, diff pane and status bar:
image
image
image
image

### DIFF
--- a/Src/MainFrm.cpp
+++ b/Src/MainFrm.cpp
@@ -223,6 +223,7 @@ BEGIN_MESSAGE_MAP(CMainFrame, CMDIFrameWnd)
 	ON_WM_ACTIVATEAPP()
 	ON_WM_NCCALCSIZE()
 	ON_WM_SIZE()
+	ON_WM_SYSCOMMAND()
 	ON_UPDATE_COMMAND_UI_RANGE(CMenuBar::FIRST_MENUID, CMenuBar::FIRST_MENUID + 10, OnUpdateMenuBarMenuItem)
 	// [File] menu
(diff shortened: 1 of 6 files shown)


sdottaka:
Thank you for the PR.
When the menu bar is hidden, pressing Alt+F etc. displays the menu, but it feels a bit strange.
Is it possible to make the menu bar appear when you press the Alt key like in other applications?
Also, I'm being very careful about adding shortcut keys these days.
I'd like to avoid adding shortcut keys if possible.

lededev:
image

sdottaka:
This PR has been merged. Thank you for the PR!
I plan to make the following modifications to this PR.

- Change the "Main Menu" menu item to "Menu Bar" to match the other menu items.
- Change the registry setting name from "Settings/HideMainMenu" to "Settings/ShowMenubar".
- Move the location of method definitions and change some method names.
- When the menu bar is set to hidden, pressing Alt+F will display the menu bar, but pressing the right key multiple times in that state will cause the menu bar to hide and then reappear, which is an unnatural behavior, so we will improve this.

PR #2511, opened 2024-10-31
Adjust right-click deselect from down to up to preserve RB + scroll anywhere

lededev:
Right button down + wheel scrolling now works from anywhere in the edit view without needing to move the cursor to the current diff line. The right-click deselect feature for diff lines is still functional, but the deselect action is now triggered on button release instead of button press.


sdottaka:
Thank you!

PR #2532, opened 2024-11-10
Press 1-3 to copy current diff line to clipbd while popup menu is open

lededev:
While right hand on mice and left hand on keyboard, press 1 .. 3 copy the current diff line when popup menu is open, no need to close menu and press Ctrl+1 .. 3, nor moving mice to menu item line then select.


sdottaka:
Thank you!

PR #2545, opened 2024-11-17
Reduce the contrast of inactive title bar text

lededev:
image


lededev:
I found a problem when AccentColorInactive bg: 0x010101, CTitleBarHelper::DrawButtons() is not using GetBackColor()'s value draw button , that make close button invisable.
But that is nothing about this PR.

sdottaka:
Thank you!

sdottaka:
>I found a problem when AccentColorInactive bg: 0x010101, CTitleBarHelper::DrawButtons() is not using getTextColor()'s value draw button , that make close button invisable.
Fixed in commit 861ee28eecf5535efd22a0d1edcb3bce1fdd147a

PR #2547, opened 2024-11-18
System Icon turn gray when app inactive

lededev:
image
Visual Studio 2022 have this feature, may usable when enable 'Use custom system color'.


lededev:
I find another problem when enable 'Use custom system color', and set Color Scheme to 'Solarized Dark', title bar background is not apply 3Dface.

sdottaka:
Thanks for the PR...
But most applications don't gray their icons when inactive. Is this important?

My concern with this is the increased size of the executable due to the icon. I want to keep WinMerge as small and fast to start as possible. (I don't think this PR will actually change the behavior much, but I want to avoid this kind of stuff accumulating.)

Is it possible to dynamically grayscale the icon instead of adding an icon?

lededev:
I've changed to lazy loading icon to improve the program startup speed, and the grayscale icon to calculate generation, removed Merge_gray.ico from resouce.

lededev:
> Thanks for the PR... But most applications don't gray their icons when inactive. Is this important?

When the interface uses a completely black skin and multiple instances are on different monitors or far apart, it is a bit difficult to determine which instance has the focus. At this time, it is faster to observe whether the system icon in the upper left corner is orange color.

sdottaka:
Thank you for the explanation.

I have merged this PR.

I will also make the following changes:

- Delete the icon in the destructor of CTitleBarHelper

- Lazy load the gray icon so that it does not need to be created until the application is inactive

lededev:
I located the switch, but since I haven't used the English version of Windows 11, I can't link it to colorPrevalence. However, it seems overly restrictive to change system settings just to enable title bar skin. Wouldn't it be simpler to allow 'Use custom system color' for this purpose? Other applications with custom title bars don't impose such limitations.
image

sdottaka:
>  I can't link it to colorPrevalence.

When "在标题栏和窗口边框上显示强调色" is turned on, I thought that HKCU\Software\Microsoft\Windows\DWM\ColorPrevelance would be set to 1, but does this actually not happen?

image

sdottaka:
> Other applications with custom title bars don't impose such limitations.
Edge respects the "Show accent color on title bars and window borders" setting.

lededev:
> When "在标题栏和窗口边框上显示强调色" is turned on, I thought that HKCU\Software\Microsoft\Windows\DWM\ColorPrevelance would be set to 1, but does this actually not happen?

I hadn't initially linked colorPrevalence with "在标题栏和窗口边框上显示强调色," but after reading some documentation, I realized the connection.

> Edge respects the "Show accent color on title bars and window borders" setting.

Some users might prefer to retain the original colorPrevalence look in Edge while trying the WinMerge dark theme. Requiring system-wide changes for this feels too restrictive.

lededev:
Perhaps making changes to the default system behavior in a new version might lead to user pushback and resistance.
However, providing an optional, user-defined switch allows users to feel they have made the choice themselves.
This approach seems sufficient, without the need to alter the entire system settings, which might impact other programs unfavorably.

lededev:
How about adding a checkbox option labeled 'Title Bar Using 3DFace'? This way, whether using accent color or 3DFace, both can utilize the user's custom system color.
image

lededev:
Microsoft Edge do not need turn off colorPrevalence when change skin to user coustom.
image

sdottaka:
What you want to achieve is to be able to change the title color without disabling colorPrevalence, correct?

lededev:
Another approach is to append `Settings/Title3DFace=1` to the `ColorSchemes/Midnight.ini` config line:
```
Settings/WordDifferenceDeletedColor=0xde000d
Settings/WordDifferenceTextColor=0xffffffff
Settings/SysColorHookEnabled=1
Settings/Title3DFace=1
```
This defines the scheme to prefer using 3DFace color for drawing the title bar. If adding a checkbox to the settings feels too complex, this method offers a simpler solution without GUI modification. However, users can only modify the ini file to change the title bar drawing behavior.

Issue #354, opened 2020-06-04
Installation -  Internal error: Failed to expand shell folder constant "userdesktop"


CodyMathis123:
WinMerge-2.16.6-x64-Setup_no_support_for_per-user_installation.exe running under system context throws the error below. I see the use of SHGetFolderPath which is deprecate, but I don't see that in the source code for the application. See the following link regarding the method being deprecated. https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetfolderpatha 

Warning: SHGetFolderPath failed with code 0x80070002 on folder 0x8010

See the attached log file for more context. Typically, the installation will succeed on the second run. 

[WinMerge-2.16.6-x64-Setup_no_support_for_per-user_installation.exe_.log](https://github.com/WinMerge/winmerge/files/4729806/WinMerge-2.16.6-x64-Setup_no_support_for_per-user_installation.exe_.log)

sdottaka:
According to the logs, it looks like you are installing WinMerge silently using software like System Center Configuration Manager instead of the usual way.
Is this my guess correct?

When I do a silent installation with administrative privileges, I get the following log output.

~~~
2020-06-05 22:08:00.493   Log opened. (Time zone: UTC+09:00)
2020-06-05 22:08:00.493   Setup version: Inno Setup version 5.6.1 (u)
2020-06-05 22:08:00.493   Original Setup EXE: c:\dev\winmerge\Build\Releases\WinMerge-2.16.6-x64-Setup_no_support_for_per-user_installation.exe
2020-06-05 22:08:00.493   Setup command line: /SL5="$151FCA,6798723,121344,c:\dev\winmerge\Build\Releases\WinMerge-2.16.6-x64-Setup_no_support_for_per-user_installation.exe" /VERYSILENT /NOCLOSEAPPLICATIONS /NORESTARTAPPLICATIONS /NORESTART /SP- /SUPPRESSMSGBOXES /LOG=C:\tmp\WinMerge-2.16.6-x64-Setup_no_support_for_per-user_installation.exe.log
2020-06-05 22:08:00.493   Windows version: 10.0.19041  (NT platform: Yes)
2020-06-05 22:08:00.493   64-bit Windows: Yes
2020-06-05 22:08:00.493   Processor architecture: x64
2020-06-05 22:08:00.493   User privileges: Administrative
2020-06-05 22:08:00.501   64-bit install mode: Yes
[...]

Issue #1814, opened 2023-04-21
Crash on typing while search is in progress


the8thavatar:
v2.16.28.0 on Windows Server 2016 
but not v2.16.6.0 on Windows 10
Start a lengthy folder comparison and type any non-directional key quickly and the program crashes without any message. Delay before pressing a key and it can handle it.

sdottaka:
So far I have not been able to reproduce this crash.
I have tried quickly pressing keys while files and folders are displayed as shown in the following image, but have not been able to reproduce it.

![image](https://user-images.githubusercontent.com/98126/233643665-792a3ebc-d9bf-4bee-a395-585750f882f1.png)

slambillion:
Crash occurs in v2.16.44.0 on Windows 11 but not on Windows 7

sdottaka:
This issue has been fixed by PR https://github.com/WinMerge/winmerge/pull/2557

PR #2557, opened 2024-11-27
Ignore special items when finding matching item on Open Dialog Find Item

slambillion:
Fix issue #354: Crash when typing a character before folder compare has finished
Special item (e.g. parent directory) does not have a filename in some versions of windows


slambillion:
Fixes issue #1814

PR #2572, opened 2024-12-10
IME switch not toggle menu when menu hide

lededev:
When the menu bar is hidden and the IME is switched using the Alt+Shift system shortcut to type some characters, the menu unexpectedly appears and gains focus. However, activating the IME indicates that the user intends to input text in the editor, not to select something from the menu.


sdottaka:
Thank you!

PR #2580, opened 2024-12-18
One-Hand mice merge without horizontal wheel

lededev:
### One-Handed Mouse Merge for Mice Without Horizontal Scroll

For users with mice that only have a vertical scroll wheel, you can still perform one-handed merge actions.

### **How to Use:**

-   **Merge from right to left**: Hold down the right mouse button and click the left mouse button, which means `Alt+Left` merge.
-   **Merge from left to right**: Hold down the right mouse button and click the middle mouse button, which means `Alt+Right` merge.

To make the newly added code more maintainable and understandable, I refactored the `MouseHook.cpp` code, using a state matrix to comprehensively cover various situations.


sdottaka:
Hmm... isn't this too complicated?
I tried this and found it to take some getting used to. Also, it can lead to unexpected merges, like merging to the right but failing and moving on to the next difference, then merging to the left by mistake when you try to go back.

If you want to merge with just your right hand, I recommend enabling merge mode.
Check the File → Merge Mode menu item and you can navigate and merge differences with just the cursor keys.
For now, I'm hesitant to merge this PR.

lededev:
I think some users will need it, so I made it, and users who don't need it should not be affected, because this operation is not commonly used and is not easy to trigger by mistake.
And after clicking the middle button, the user still has enough time to think about whether to proceed with the merge and which direction should scroll to.

lededev:
The Microsoft PowerToys FancyZones module includes a feature similar to this:

- **Move a Window**: Hold the left mouse button, drag the window to a zone.
- **Drop in a Zone**:
  - Hold the `Shift` key to drop the window in the zone.
  - Alternatively, keep holding the left mouse button and click the right mouse button to change the state, equivalent to holding `Shift`.
- **Merge Zones**:
  - Click the middle mouse button while holding the left mouse button. This simulates holding the `Ctrl` key, allowing you to merge zones and drop the window into a larger area spanning multiple zones.

lededev:
I use a Logitech G502 mouse, which has a very firm middle button that requires significant pressure to activate. If your mouse's middle button is more sensitive, it might accidentally trigger while scrolling. As an alternative, I can modify this operation to be triggered by the left button instead. If switched to the left button, users will need to move their finger away from the scroll wheel to change the state and then return to the scroll wheel position. This greatly reduces the chance of accidental activation, but also lowers efficiency.

lededev:
@sdottaka I've simplified the operation: holding the right mouse button and clicking the left button is equivalent to Alt+Left, while holding the right mouse button and clicking the middle button is equivalent to Alt+Right.

This method can be more efficient doing "merge" than using the keyboard alone in certain scenarios. For instance, when dealing with multi-line diff blocks and needing to select small block of differences in the middle of long lines to merge to the other side, using the mouse is often faster than using the keyboard. Once the selection is made by mouse, you can simply hold the right mouse button and click the left or middle button to complete all the operations with only one hand, without needing to switch to the keyboard.

sdottaka:
Thank you for updating the PR.
It's easier than the previous one. However, I don't know why, and I don't know if it will reproduce in your environment, but when I press the left or middle mouse button while pressing the right mouse button in the folder comparison window, WinMerge seems to hang up.

Also, I think that having six different ways of operating one function is too much. Sorry, but I will reject this PR.

Also, when copying to the right with Alt + Shift + Wheel Down, an unexpected IME switch occurs if you release the Alt key first.
Regarding this, I am thinking of deleting Alt + Shift + Wheel Down/Wheel Up.

lededev:
Unfortunately you're not going to accept this change, but I'll continue to use the software to do my job, maybe maintaining a version for my own personal use until a better cross-platform alternative emerges.

PR #2601, opened 2025-01-05
Update Brazilian.po

Marcellomco:
Added translation for "Enable per-file plugin application with file number specification (#2598)"


sdottaka:
Thank you!

PR #2623, opened 2025-01-23
Update Brazilian.po

Marcellomco:
Added translation for "Add a Prompt to Copy Only Differences in Folder Comparison (#2622)"


sdottaka:
Thank you!

PR #2634, opened 2025-02-02
Add Visual Studio Dark color scheme

rkodey:
Add new color scheme based on the Visual Studio theme in vscode.  It seems a few users have given the thumbs-up, so I thought I'd contribute.  Thanks!


sdottaka:
Thank you for the PR!
I'm concerned about the inclusion of the word Visual Studio, so I think I'll change the name to something like VS Dark.ini.

sdottaka:
I renamed the file to VS Dark.ini.
If you have any suggestions for a different name that doesn't include Visual Studio, please let me know.

PR #2646, opened 2025-02-13
Use zopflipng to reduce size of splash.png

CharlesMengCA:
Maybe you can also remove anything related to splash.jpg

 <Image Include="res\splash.jpg" />


sdottaka:
Thank you!

PR #2655, opened 2025-02-23
Modify the "Options (Compare > Image)" dialog.

tjmprm77:
The "Defaults" button now resets the following settings to their default values:
- Enable image compare in folder compare
- OCR result
dialog


sdottaka:
Thank you!

PR #2679, opened 2025-03-18
Hans Eriksson updated the Swedish translation

gerundt:
Hans Eriksson sended me a updated Swedish translation


sdottaka:
Thank you!

PR #2694, opened 2025-03-25
Twilight - new dark theme

dcog989:


dcog989:
@sdottaka - if you can help fill in any of the `??????` in comments, I will update the source so that it is fully commented and makes a good template for others.

sdottaka:
Thank you!
This PR has been merged.
I will fix the ??? part later. (If the English is wrong in the ??? part, I would appreciate it if you could make another PR.)
The Custom Colors is the part shown in the attached image, so I think you probably defined it. Therefore, I don't know the name.
image

Also, comments in INI files cannot be written at the end of the line. Comments are considered to be values. (This INI file worked because all the values ​​were integers. If the values ​​were strings, the comments would be considered to be values.)
For this reason, I will move the comments at the end of the lines to the beginning of the lines later.

dcog989:
Thank *you*! 

I will check any updates for grammar, etc and raise new PR as necessary.

I see. So the `Custom Colors` are not used by the theme? They simply show what is in the color picker? Then the `Custom Colors/0`, etc could be deleted from the theme?

Thanks for fixing the comments. 

P.S. I just noticed I left `#7e4242` in comments - that can be deleted if you are moving comments.

sdottaka:
> I see. So the Custom Colors are not used by the theme? They simply show what is in the color picker? Then the Custom Colors/0, etc could be deleted from the theme?

Yes, they are not used and can be deleted.

I have committed the corrections, including comments.

PR #2726, opened 2025-04-13
Update Brazilian.po

Marcellomco:
Added translation for "Add Output Pane for Logs and Messages (#2663) (2)"


sdottaka:
Thank you!

PR #2727, opened 2025-04-14
Update Finnish.po

6sto:
This pull request updates the Finnish localization file.

Changes include:
* Translated missing strings.
* Reviewed existing translations for accuracy and corrected errors.

### DIFF
--- a/Translations/WinMerge/Finnish.po
+++ b/Translations/WinMerge/Finnish.po
@@ -3,6 +3,7 @@
 #
 # Translators:
 # * Veikko
+# * 6sto
 #
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you for your PR!
However, not only the Finnish text but also the original English text (msgid) has been modified, as shown below.

https://github.com/WinMerge/winmerge/pull/2727/files#diff-e6232560c1c602cad36e184ac4d07489cf9a0ae032355133e1db28b4ec83f36fL286-R286

Is this intentional? If you change msgid to something outside of WinMerge's control, some translations in Finnish.po will be lost when WinMerge is rebuilt, as shown below, so could you please change msgid back to its original state?

image


If you want to modify msgid, you will need to modify not only Finnish.po but many other files as well, so please make a separate PR.

6sto:
Sorry. That was not my intention. I'll recheck everything.

6sto:
I have to check this again. "I'll be back!"

Issue #1423, opened 2022-07-26
Open multiple selected files in registered or external editor


sdottaka:
### Discussed in https://github.com/WinMerge/winmerge/discussions/1422

<div type='discussions-op-text'>

<sup>Originally posted by **ryandenki** July 26, 2022</sup>
Currently when selecting multiple files and right-clicking to bring up the context menu, the "Open ..." options are greyed out.

This means you cannot open a selected set of files in the editor together, but must deselect all files, then individually one-by-one open the context menu and select the appropriate "Open ..." option each time.

It would be helpful to be able to open multiple files in an editor, perhaps by iterating over the list and opening each one separately.

</div>

PR #2749, opened 2025-04-27
Add the feature to open multiple files in the folder compare window requested in #1423.

tjmprm77:
The following menu items are now available in the context menu when multiple files or folders are selected:
- "Open Left/Middle/Right" > "With Registered Application"
- "Open Left/Middle/Right" > "With External Editor"
- "Open Left/Middle/Right" > "Open Parent Folder..."

("Open Left/Center/Right" > "With..." are not included in this feature addition because multiple "Open With" dialogs cannot be opened at the same time at least in Windows 11.)


sdottaka:
Thank you!

PR #2751, opened 2025-04-28
Fix syntax warning in `GetTranslationsStatus.py`

gerundt:
Since [Python 3.12](https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes) the script `GetTranslationsStatus.py` shows an *syntax warning* because of a *invalid escape sequence*.

This PR fix this.


sdottaka:
Thank you!

PR #2769, opened 2025-05-02
Use themed size grip, use toolbar's tooltips instead of controls bar's

ozone10:
This PR is mainly for preparation for dark mode.

Before
image
image

After
image
image


sdottaka:
Thank you!

Issue #2702, opened 2025-03-31
[BUG} Phantom line added when identical blocks of text pasted


dcog989:
Reproduce:

1. New Document
2. paste a multi-line block of text in 1st pane
3. focus 2nd pane by clicking approx. centre
4. paste identical block in 2nd pane 
5. result = a 'phantom' line is added to the 2nd pane

![Image](https://github.com/user-attachments/assets/38e34a13-9850-4941-b2d0-d1c0bca16f26)

Note: problem does not occur if 2nd pane is focused by clicking on top or bottom line.

Version: 2.16.47b

sdottaka:
Hmm I haven't been able to reproduce this issue so far...

dcog989:
It happens every time for me - with any block of text. Maybe try my settings? 

[Winmerge-registry-20250331-174417.zip](https://github.com/user-attachments/files/19542999/Winmerge-registry-20250331-174417.zip)

sdottaka:
This issue has been fixed in commit [bd281a8](https://github.com/WinMerge/winmerge/commit/bd281a837450a54c3a657e37c5ae53e6030561c9). Thank you for your report

dcog989:
@sdottaka - would you consider creating a nightly branch so we can test changes before the production release?

sdottaka:
If it is an x64 binary, there is a CI Build artifact.
You can download it from the following URL.
Note that these cannot be code signed, so Windows may block their execution.

https://ci.appveyor.com/project/sdottaka/winmerge/build/artifacts
https://github.com/WinMerge/winmerge/actions/runs/14254676436

dcog989:
Excellent! I'll keep an eye on those in future. Thank you!

dcog989:
I just tried [winmerge-2.16.47-9c4970d98-x64-exe.zip](https://ci.appveyor.com/api/buildjobs/2v1b2o7a2s6unlmo/artifacts/build%2FReleases%2Fwinmerge-2.16.47-9c4970d98-x64-exe.zip). The original problem is fixed in some cases, but not others. Results:

## Case 1 

If I paste content with length >50% of viewport that ends with a blank line anywhere other than the top line in pane 2, then the content in pane 2 is scrolled out of the view window:

![Image](https://github.com/user-attachments/assets/4a016b99-f7ab-4c0b-8ffe-367e689fd6ff)

## Case 2

If I paste content *without* a blank line at end, then the original problem occurs:

[...]

PR #2785, opened 2025-05-19
Force text to paste on line 1 rather than 'void' space

dcog989:
This addresses #2702 by forcing paste in 2nd or 3rd pane to begin on line 1 of text input, rather than below it in the 'void' area which then causes false positive difference.

I have built and tested locally.

### DIFF
--- a/Src/MergeEditView.cpp
+++ b/Src/MergeEditView.cpp
@@ -1268,12 +1268,56 @@ void CMergeEditView::OnEditPaste()
 /**
  * @brief Called when "Paste" item is updated
  */
-void CMergeEditView::OnUpdateEditPaste(CCmdUI* pCmdUI)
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you for the PR.
However, it introduces a compile error because OnEditPaste is defined twice.
Also, the code always pastes text at line 1, regardless of the caret position.
For example, if the caret is on line 10, the pasted text still goes to line 1.
Unfortunately, we cannot accept this PR.

sdottaka:
There have been various issues in the past with inserting or pasting text into ghost lines, and it's not a simple problem. I plan to look into it after finishing the feature I'm currently working on.

dcog989:
OK, thanks for taking the time with it. I understand there is more complexity than I'm aware of. It's a small issue in the end.

PR #2812, opened 2025-06-12
Tamil translations reviewed and updated few messages

TamilNeram:
Updated both Tamil files for missing entries and reviewed few messages.


sdottaka:
Thank you!

PR #2821, opened 2025-06-24
[Enhancement] Tab Switching and Tab Bar Scrolling with Mouse Wheel

anthonyleestark:
## ✨ What's New in This Enhancement
This PR enhances the tab control behavior by enabling intuitive **tab switching** using the mouse wheel with `Ctrl` or `Shift` modifiers, and also supports **tab bar scrolling**.
- Switch tabs with the mouse wheel:
  - `Ctrl + MouseWheel (Up/Down)`: Switches to the previous/next tab **WITH wraparound**
  - `Shift + MouseWheel (Up/Down)`: Switches to the previous/next tab **WITHOUT wraparound**
  - `Ctrl + Shift + MouseWheel (Up/Down)`: Switches to the first/last tab
- Support Tab bar scrolling: When **no modifier key is held** and the tabs exceed the visible area, rotating the mouse wheel scrolls the tab bar.

## ✅ Test Plan
- Manual testing with `Ctrl + MouseWheel`: cycles tabs forward/backward with wraparound.
- Manual testing with `Shift + MouseWheel`: moves forward/backward without wrapping.
- Manual testing with `Ctrl + Shift + MouseWheel`: switches to the first/last tab
- Confirmed tab bar scrolls only when necessary

## 📋 Checklist
 - [x] No regressions in tab scrolling behavior
 - [x] Modifier key handling confirmed across input cases

### DIFF
--- a/Src/Common/MDITabBar.cpp
+++ b/Src/Common/MDITabBar.cpp
@@ -93,6 +93,25 @@ void CMyTabCtrl::SetActive(bool bActive)
 	m_bActive = bActive;
 }
 
+/**
+ * @brief Activate the specific tab by index.
+ * @param nTabIndex [in] Tab index to activate
+ */
+void CMyTabCtrl::ActivateTab(int nTabIndex)
+{
+	if (nTabIndex < 0 || nTabIndex >= GetItemCount())
+		return;
+
+	SetCurSel(nTabIndex);
+
+	// Notify tab selection changed
+	NMHDR nmhdr = {0};
+	nmhdr.hwndFrom = GetSafeHwnd();
+	nmhdr.idFrom = GetDlgCtrlID();
+	nmhdr.code = TCN_SELCHANGE;
+	GetParent()->SendMessage(WM_NOTIFY, nmhdr.idFrom, reinterpret_cast<LPARAM>(&nmhdr));
+}
+
 static inline COLORREF getTextColor()
 {
 	return GetSysColor(COLOR_WINDOWTEXT);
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you for the PR.

We aim to keep WinMerge's mouse and keyboard interactions consistent with other applications whenever possible.

While switching tabs with the mouse wheel is supported in some editors like Visual Studio Code (e.g., Shift + Mouse Wheel Up/Down), it's not a widely adopted convention across most applications.

We propose adopting only the Shift + Mouse Wheel Up/Down behavior for now, following the VS Code precedent.

We'd prefer to reserve Ctrl(+Shift) + Mouse Wheel for potential future use, depending on broader UI trends and user expectations.

anthonyleestark:
> We propose adopting only the Shift + Mouse Wheel Up/Down behavior for now, following the VS Code precedent.
We'd prefer to reserve Ctrl(+Shift) + Mouse Wheel for potential future use, depending on broader UI trends and user expectations.

@sdottaka 
Thank you for your response. I’ll go ahead and remove the Ctrl(+Shift) + Mouse Wheel functionality from this PR, and keep only the Shift + Mouse Wheel and tab bar scrolling for now. We’ll gather user feedback and consider bringing it back later if needed.

sdottaka:
Thank you for updating the PR.
I tested it, and while Shift + Wheel Up/Down switches the tabs, the file or folder comparison windows don’t change accordingly.
Does the switch work properly in your environment?

anthonyleestark:
> the file or folder comparison windows don’t change accordingly.

@sdottaka 
Ah, sorry, it's my mistake.
I initially created a separate function to activate the tab, and it worked correctly.
However, I accidentally removed it during a code refactor, which left only the SetCurSel call without triggering the TCN_SELCHANGE notification. I later re-tested using blank tabs and didn’t notice the bug at that time.
I've updated the fix to restore the tab activation function.

sdottaka:
Thank you!

Issue #324, opened 2020-04-26
Feature Request: Allow comparing multiple files in one action


vertigo220:
When having a large number of files (in this case, images) that are different, instead of having to go one-by-one and double-click on them, observe the visual differences, close the compare tab, and move on to the next one, it would be preferable to be able to select multiple files, right-click, select "Compare" (which is currently grayed out when multiple files are selected), and have them open in separate tabs, which would then allow the first one to be selected, visually compared, and closed, at which point the next tab would come in focus, allowing for a quick comparison and click to close, comparison and click to close, and so on.

blackboxlogic:
Until this enhancement is added to this project, here's a workaround for anyone that's stuck: https://sourceforge.net/projects/diffuse/

Gitoffthelawn:
@blackboxlogic The main page for diffuse states that it is a "tool for comparing and merging text files".  Does it compare images as well?

blackboxlogic:
@Gitoffthelawn Just tried, the tool I suggested compares images as if they were text file, so... No.

sdottaka:
Closing this issue as it was implemented by PR #2825.

PR #2825, opened 2025-06-29
Add the feature to compare multiple files in one action in the folder compare window requested in #324.

tjmprm77:
Add the feature to compare multiple files in one action in the folder compare window requested in #324.

The following menu items are now available in the menu bar or the context menu when multiple files or folders are selected:
- Menu bar
  - "Merge" > "Compare"
- Context menu
  - "Compare"
  - "Compare in New Window"
  - "Compare As" > "Text"
  - "Compare As" > "Table"
  - "Compare As" > "Binary"
  - "Compare As" > "Image"
  - "Compare As" > "Webpage"

WinMerge already has the feature to compare two or three items where only one file exists each as you know, and this feature takes precedence over the feature added in this PR.
This means that the comparison process added in this PR will only be used in cases where the existing process is not available.

For example:
  - File A: Only exists in the left side.
  - File B: Only exists in the right side.
  - File C: Both Exists in the left and right sides.

  1. Select File A and File B and compare them.
      This will open the following file compare window:
      - File A (Left) vs. File B (Right)

  2. Select File A and File C and compare them.
      This will open the following file compare window:
      - File A (Left) vs. Empty File
      - File C (Left) vs. File C (Right)

  3. Select File A, File B and File C and compare them.
      This will open the following file compare window:
      - File A (Left) vs. Empty File
      - Empty File vs. File B (Right)
      - File C (Left) vs. File C (Right)

### DIFF
--- a/Src/DirView.cpp
+++ b/Src/DirView.cpp
@@ -715,6 +715,7 @@ static void NTAPI CheckContextMenu(BCMenu *pPopup, UINT uIDItem, BOOL bCheck)
 
 /**
  * @brief User right-clicked in listview rows
+@* @param [in] point Point where the mouse was right-clicked.
  */
 void CDirView::ListContextMenu(CPoint point, int /*i*/)
 {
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you!

PR #2827, opened 2025-06-30
[Enhancement] Enable/disable automatic navigation to a line when clicking on the Location Panel

### LINKED ISSUE #2811: Scrolling with location pane loses cursor position
When you scroll with the scroll bar the cursor is not affected.
When you scroll with the side location pane, the cursor position is lost (unless there's an option which I didn't spot). This causes a problem if you are scrolling to a new position with the pane in order to highlight an area for committing across, whereas I can see no advantage in the cursor moving since even if you wanted the cursor in the new position the scrolling is not exact such that it will fall in the right place except by luck. It's not entirely clear to me what formula is actually determining what the new cursor location becomes.
I can see though that scrolling with the scroll bar does the same movement but keeps the cursor but in my view that's not intuitive to the user who will often want to drag the location scroller, so my recommendation would be to maintain the cursor position when scrolling with the location pane.

Cheers! David

anthonyleestark:
## 🐞Description
The issue was reported in #2811

## ✨ What's New in This Enhancement
This PR implements a new option in the Location Panel’s context menu that enables or disables automatic navigation to a line when clicking on the Location Panel.

<div align="center">

</div>

This option is enabled by default, preserving the current behavior.
When disabled, clicking on the Location Panel will not move the cursor to the corresponding line.

## ✅ Test Plan
Manual testing on Windows 10/11

## 🔗 Related Issue
Merging this PR can close #2811

## 📋 Checklist
- [x] No regressions observed


sdottaka:
We’ll accept this PR as it adds a useful option while keeping the default behavior.

However, if more options are added directly to context menus, they may become too cluttered.
In the future, we might move such options to a settings window or reorganize the menus.

Also, would it be okay to rename “Go to Line on Click” to “Move Cursor on Click” for clarity?

anthonyleestark:
@sdottaka 
> would it be okay to rename “Go to Line on Click” to “Move Cursor on Click” for clarity?

I've updated it

sdottaka:
Thank you!

Issue #2738, opened 2025-04-17
maximised tabs' clickable part dont extend to screen edge


tsehygit:
similar to #2520 but for tabs' clickable part. clicking the edge dont switch to that tab. trying to move tabs remove window from maximised size and drags it instead.

sdottaka:
You’re probably referring to clicking on the very top edge of the title bar — is that correct?

Since double-clicking the top edge of the title bar exits maximized mode, this behavior has some merit.
However, browsers like Edge and Firefox do switch tabs when clicking the top edge of the title bar, so perhaps WinMerge should follow that behavior too...

tsehygit:
>You’re probably referring to clicking on the very top edge of the title bar — is that correct?

exactly.

karbun01:
I believe I found a solution to your issue. Check it out here [click](https://forreststonesolutions.com/robots/)

anthonyleestark:
@tsehygit @sdottaka 

I made a PR which maybe considered as a temporary solution for this issue.

> Since double-clicking the top edge of the title bar exits maximized mode, this behavior has some merit.

Since you said so, I didn't intercept this behavior.
With this PR, the new behavior will be both selecting the tab and exitting maximized mode.

> However, browsers like Edge and Firefox do switch tabs when clicking the top edge of the title bar, so perhaps WinMerge should follow that behavior too...

I've noticed that browsers render the tab button at maximum height when in maximized mode.
If we aim to follow their behavior, I think we should apply the same tab button rendering mechanism in WinMerge.
I actually also tried rendering the tab button at maximum height, but it didn’t look great. To improve the appearance, we should render the bottom border of the tab button the same way browsers do.

Anyway, my PR isn't complete yet. Please test it and provide some suggestions to help me finish it.

sdottaka:
This issue has been resolved by PR #2828.

PR #2828, opened 2025-07-01
Treat a mouse click on the very top edge of the title bar while maximized as if it's within the tab button

anthonyleestark:
## 🐞Description
The issue was reported in #2738
This PR implements a temporary solution for the issue.

## ✅ Test Plan
Manual testing on Windows 10/11

## 📋 Checklist
- [x] No regressions observed

### DIFF
--- a/Src/Common/MDITabBar.cpp
+++ b/Src/Common/MDITabBar.cpp
@@ -711,6 +711,7 @@ void CMyTabCtrl::UpdateToolTips(int nTabItemIndex)
 BOOL CMDITabBar::Update(bool bOnTitleBar, bool bMaximized)
 {
 	m_bOnTitleBar = bOnTitleBar;
+	m_bMaximized = bMaximized;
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you for the PR. However, the fix feels a bit complex. I’ll take a look later to see if adjusting the tab control position alone could resolve it.

anthonyleestark:
@sdottaka 

> Thank you for the PR. However, the fix feels a bit complex

As I said, this is just a temporary solution for the issue. 
This PR is open for discussion; it's completely okay not to be merged.

Actually, I didn't try adjusting the tab control size and position, only tried adjusting the tab button size when rendering, and it didn't look very good. A better way would be to render the tab buttons the same shape as browsers like Edge, Chrome, or Firefox.

sdottaka:
I’ll go ahead and merge it for now, and then make some adjustments on top of it to simplify the logic a bit. Thank you again for the contribution!

PR #2829, opened 2025-07-02
Update Brazilian.po

Marcellomco:
Added translation for "Enable/disable automatic navigation to a line when clicking on the Location Panel"


sdottaka:
Thank you!

PR #2849, opened 2025-07-22
Update Brazilian.po

Marcellomco:
Added translation for "Improve filter system: expression support, pf: prefix, and UI enhancements"


sdottaka:
Thank you!

PR #2850, opened 2025-07-22
Italian language update

bovirus:
@sdottaka 

Please merge. Thanks.

Please remove from .po in the strings use of "\r" that is not standard for gettext tool ("\n" is accepted by getetxt tool).
Threre 54 strings with "\r\n".

If you can formatted po file with long line width (400 chars) to avoid strange new line after 77/80 chars.

### DIFF
--- a/Translations/WinMerge/Italian.po
+++ b/Translations/WinMerge/Italian.po
@@ -10,17 +10,17 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: WinMerge (Italian) 02.07.2025\n"
+"Project-Id-Version: WinMerge (Italian) 22.07.2025\n"
 "Report-Msgid-Bugs-To: https://github.com/WinMerge/winmerge/issues/\n"
 "POT-Creation-Date: 2025-04-17 09:05+0000\n"
-"PO-Revision-Date: 2025-07-02 21:30+0200\n"
+"PO-Revision-Date: 2025-07-22 18:26+0200\n"
 "Last-Translator: bovirus <bovirus@gmail.comn>\n"
 "Language-Team: bovirus (bovirus@gmail.com)\n"
 "Language: it\n"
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you!

We use \r\n because otherwise line breaks may not display correctly in message boxes, and generated files might use LF instead of the standard CR+LF line endings on Windows. Since this affects behavior significantly, we would prefer to keep it as is.

WinMerge's translation tools do not control line wrapping in .po files. Could it be your PO editor that is automatically inserting line breaks?

bovirus:
@sdottaka

About forced new line could be the tool (Poedit) that add forced new line.

About "\r\n" there 54 strings with these end line and about 350 strings with only "\".
Why only for these 54 strings (27 source+27 translated) with "\r\n"?

sdottaka:
We intentionally use \r\n in those 54 strings because they are used in message boxes or output files where Windows requires CR+LF line endings for proper formatting.
The other strings don’t need special line endings, so they just use \n.
So \r\n is used only where it’s necessary for correct behavior on Windows.

If we use \n instead of \r\n, the message appears as shown in the first image.
It should be displayed as shown in the second image.

You’ve replaced \r\n with \n, but this clearly causes a problem.
Would it be okay if I revert your change to restore the correct behavior?

image
image

bovirus:
@sdottaka 

For me it's ok if you revert change of "\n" in "r\n" where it needs..
Did you try to use "\n\n" instead "\r\n"?

What are the menu steps in WinMerge the scren that you posted as exemple?

sdottaka:
>Did you try to use "\n\n" instead "\r\n"?

I tried using \n\n earlier, but it had no effect.

>What are the menu steps in WinMerge the scren that you posted as example?

In the File Compare window, choose Plugins → Edit with Prediffer, then select the SelectLines plugin.

Issue #2852, opened 2025-07-23
WinMerge - Stngs untranslated


bovirus:
@sdottaka 

<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/11eae9b1-b8bc-499f-bfd5-67fc65aa5afa" />

<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/b053970f-d612-4ae7-84e4-64c13747a10a" />

<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/b053d1d9-4e59-4f3f-83ff-5a4d3ecbc0e8" />

The same strings "Disabled" appera in different context where in not english language could be male or female (different translation).

<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/13e1d230-2eff-47ad-9a5d-281224164462" />
Why option names start with lowercase?

<img width="677" height="488" alt="Image" src="https://github.com/user-attachments/assets/89c0a814-fa01-4725-8e6a-1f593310a7ad" />
<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/76280f24-550b-41b9-b4e6-b0afcb244d2e" />
<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/7db6ce23-ecd7-444a-9881-15ba8e1214d7" />
Panels with strings untranslated

<img width="678" height="483" alt="Image" src="https://github.com/user-attachments/assets/b22aee60-35d0-48fd-ae6d-15c224799d7a" />
Strings untranslated

<img width="551" height="412" alt="Image" src="https://github.com/user-attachments/assets/e08e1611-2eca-4b66-9267-16b98b0a3d71" />
String untranslated

sdottaka:
> The same strings "Disabled" appera in different context where in not english language could be male or female (different translation).

I see. This might be a bit tricky to handle.

> Why option names start with lowercase?

This comes from the diff algorithm used in the diff library (libxdiff), which follows that naming convention.

> Panels with strings untranslated

These translations are managed in the subrepository frhed, so please submit them here:
https://github.com/WinMerge/frhed/tree/github/3way-diff-gplv2/Translations/Frhed

> String untranslated

As far as we know, copyright notices are generally not translated.

bovirus:
@sdottaka 

About
[...]

PR #2862, opened 2025-07-26
Update Brazilian.po

Marcellomco:
Added translation for 'Make "Disabled" entries in General options separately translatable (refs #2852)'


sdottaka:
Thank you!

PR #2869, opened 2025-07-28
Readme Inno Setup script improvements

bovirus:
@sdottaka 

Please check and merge. Thanks.

### DIFF
--- a/Installer/InnoSetup/WinMergeX64.iss
+++ b/Installer/InnoSetup/WinMergeX64.iss
@@ -76,6 +76,9 @@ DirExistsWarning=no
 ;Tells the installer to only display a select language dialog if the an exact match wasn't found
 ShowLanguageDialog=auto
 
+WizardStyle=modern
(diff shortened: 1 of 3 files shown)


sdottaka:
Thanks for the PR. I’ve merged it, but just a note:
WizardStyle=modern and WizardResizable=yes don’t work with the older version of Inno Setup 5.x we're currently using.
We’re using the older version because newer ones have caused false positives with some antivirus vendors in the past.

So I’ll revert those two settings. Also, just a note that the installer scripts include more than just WinMergeX64.iss and WinMergeX86.iss.

bovirus:
@sdottaka 

Sorry about WizarDStyle but in Inno Setup command reference was not explained as specific for Inno Setup 6.x.

The installer is multilanguage and the way to show README it's equal for all languages.
Then I propose to modify the README for other language as I changed the italian version of README.
New line only to separate any paragraph.

PR #2914, opened 2025-08-25
Turkish translation support

bilimiyorum:
New string entries
Some translation corrections (e.g. correction of overlapping text due to long translation) 
Punctuation and spelling corrections


sdottaka:
Thank you!

PR #2951, opened 2025-09-21
Update Brazilian.po

Marcellomco:
Added translation for "Add option to ignore line breaks (treat as spaces)"


sdottaka:
Thank you!

PR #2954, opened 2025-09-22
Ukrainian PO translation was synced with English (autogenerated changes)

YevgenBorodkin:
This change contains only autogenerated changes that were introduced by PoEdit after syncing with original english strings.


sdottaka:
Thank you!

PR #2956, opened 2025-09-24
Italian language update

bovirus:
@sdottaka 

Please merge. Thanks.

Note: recent strings adding seems about strings that are very smilar (except for uppercase/lowercase)

Ignore line breaks (treat as spaces)
Ignore Line Breaks (Treat as Spaces)

Is it necessary two strings?

### DIFF
--- a/Translations/WinMerge/Italian.po
+++ b/Translations/WinMerge/Italian.po
@@ -10,10 +10,10 @@
 #
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you!

>Ignore line breaks (treat as spaces)
>Ignore Line Breaks (Treat as Spaces)

The first string is used in the Options window, while the second one is used in the menu.

PR #2977, opened 2025-10-13
Update Brazilian.po

Marcellomco:
Added translation for "Add support for prop, leftprop, middleprop, and rightprop functions in filter expressions"


sdottaka:
Thank you!

PR #2995, opened 2025-10-23
Update Brazilian.po

Marcellomco:
Added translation for "Add option for specifying additional comparison conditions (#2963)"


sdottaka:
Thank you!

PR #3002, opened 2025-10-24
Add Ctrl+PageUp/Ctrl+PageDown as alternative hotkeys to Ctrl+Tab/Ctrl+Shift+Tab

### LINKED ISSUE #2942: [Feature request] Alternate hotkeys for tab navigation
Opening the previous/next tab using a keyboard is only possible via Ctrl+Shift+Tab/Ctrl+Tab.

I'd like a secondary key combo to be added, which is Ctrl+PageDown/Ctrl+PageUp

Many programs have that exact secondary combo, for example Chrome, Firefox, Notepad++.

ethanhaines:
Hello, I am a student who uses WinMerge and this is one of my first open source contributions that I've done on my own time, apologies if I have done something incorrect. I saw issue #2942 and realize I use the same hotkeys in browsers, which gave me motivation to try and tackle this!

I made my changes in the same place that the ctrl+tab hotkeys are checked in mainfrm.cpp, then I just had to add logic in windowsmanagerdialog.cpp to also handle the new hotkeys. I then had to change the hexmergeview.cpp because the hotkeys were not working in the binary editor after testing.

I built and tested the changes on my Windows 11 computer and the behavior exactly matched ctrl+tab and ctrl+shift+tab usage.

Closes #2942

### DIFF
--- a/Src/HexMergeView.cpp
+++ b/Src/HexMergeView.cpp
@@ -544,6 +544,11 @@ BOOL CHexMergeView::PreTranslateMessage(MSG* pMsg)
 			return false;
 		}
 	}
+
+	if (GetAsyncKeyState(VK_CONTROL) < 0 && (pMsg->wParam == VK_PRIOR || pMsg->wParam == VK_NEXT))
+	{
+		return false;
+	}
 	return m_pif->translate_accelerator(pMsg);
 }
 
--- a/Src/MainFrm.cpp
+++ b/Src/MainFrm.cpp
@@ -2224,6 +2224,13 @@ BOOL CMainFrame::PreTranslateMessage(MSG* pMsg)
 		return TRUE;
 	}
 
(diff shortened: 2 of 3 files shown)


sdottaka:
Thank you for the PR and your contribution!
However, I feel that Ctrl+Page Down/Up should switch windows directly like Ctrl+F6, without showing the window switch dialog as Ctrl+Esc does. This behavior would match Visual Studio and VS Code.
Additionally, to allow for possible future customization of keyboard shortcuts, I prefer implementing this via a Win32 accelerator resource rather than hardcoded logic.
This change has been made in commit d0fd205.
I appreciate your effort on the PR, but unfortunately we cannot merge it as-is.

ethanhaines:
Okay, thank you for the feedback! This was still a valuable learning experience.

PR #3011, opened 2025-10-28
i18n: Improve Traditional Chinese Terminology

love80312:
Standardized key technical terms to ensure consistency and better local usage


sdottaka:
Thank you!

PR #3013, opened 2025-10-29
Added timestamps

Thorium:
Small change: 2 todos fixed


sdottaka:
Thanks for cleaning up the TODO comment.
However, since tFiles.ltime and tFiles.rtime are currently unused, assigning timestamps to them seems unnecessary.
Maybe it would make more sense to remove these fields from PATCHFILES altogether.

sdottaka:
As the timestamp fields are currently unused and we don't plan to adopt this change, we’ll close the PR for now.
Thanks for your contribution.

PR #3028, opened 2025-11-12
Replace line breaks with space in MSExcel formulas

JoHoGH:
This is a suggestion to remove line breaks from formulas when unpacking MSExcel files so the formula is not broken across multiple lines in Winmerge. Line breaks are often added to improve readability in the MSExcel formula bar, but they are not of any interest when looking for differences.

### DIFF
--- a/Plugins/dlls/CompareMSExcelFiles.sct
+++ b/Plugins/dlls/CompareMSExcelFiles.sct
@@ -168,22 +168,26 @@ function writeFormulas(fo, sht) {
   var row, col, formula;
   var rowOffset = sht.UsedRange.Row;
   var colOffset = sht.UsedRange.Column;
+  var keepLineBreaks = regRead(REGKEY_PATH + "CompareFormulasKeepLineBreaks", false);
   try {
     var varCells = sht.UsedRange.Formula;
     if (typeof(varCells) === "string" || typeof(varCells) === "number" || typeof(varCells) === "date") {
-      // Replace line breaks with space
-      var cleanFormula = String(varCells).replace(/[\r\n]+/g, " ");
-      fo.WriteLine(getAddr(1 + rowOffset - 1, 1 + colOffset - 1) + ": " + cleanFormula);
+      var outFormula = String(varCells);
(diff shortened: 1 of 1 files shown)


sdottaka:
Thank you for the PR.
In the table comparison window, enabling View → Wrap Text makes WinMerge display actual line breaks within cells.
Since this is often useful, I’d prefer not to always replace line breaks with spaces.
Could this conversion be made optional instead?

JoHoGH:
Thanks for looking! Currently, formulas with line breaks do not behave like cell values with line breaks.  In the current version any line breaks in the formula creates separate line in the table and changing "Wrap text" does not change that.
For example:
Screenshot 2025-11-12 142609  
will look like this in Winmerge table comparison (with Wrap Text off)
Screenshot 2025-11-12 141132

With my proposed change it gives this
 
Screenshot 2025-11-12 143228

JoHoGH:
Hi. I tried a different approach using the `quoteIfNeeded` function to preserve the line breaks in the formula. Code change and result are below. Personally I prefer my earlier version that changed line breaks to space
Result (Wrap text selected)
Screenshot 2025-11-12 145105
Code change 
Screenshot 2025-11-12 144920

sdottaka:
Sorry, I misunderstood — I thought you were replacing line breaks with spaces for all cell values, not just formulas.
Quoting formulas with double quotes makes them very hard to read, so I agree that’s not a good approach.

However, I still don’t think replacing line breaks should be the default. It would be better to make it configurable via an option.

JoHoGH:
Hi, I've pushed a change to add a checkbox to select whether or not to keep line breaks in the formulas. 
And thankyou for this tool - it's awesome!

sdottaka:
Sorry for the late reply, and thank you for the PR. I’ve merged it.

However, I don’t want to make replacing line breaks with spaces the default behavior. I plan to make the following adjustments later:

- Rename the option to “Replace line breaks in formulas with spaces” and set its default value to false.
- Make this option name translatable.

PR #3034, opened 2025-11-17
Update Brazilian.po

Marcellomco:
Added translation for "Replace line breaks with space in MSExcel formulas (#3028) (2)"


sdottaka:
Thank you!

PR #3064, opened 2025-12-08
Update Brazilian.po

Marcellomco:
Added translation for "Async binary check during identical message box (#3060)"


sdottaka:
Thank you!

PR #3073, opened 2025-12-12
Update ChineseTraditional translation: translate missing strings

love80312:
This line cannot be displayed correctly after translation.

msgid "Failed to extract archive.\nCompare as text file?"
此行不能正確顯示翻譯


sdottaka:
Thank you for submitting the translation. The issue where that message was not translated has been fixed in commit e15b924.

PR #3085, opened 2025-12-21
Update Brazilian.po

Marcellomco:
Added translation for "Add option to select user data location (AppData or Documents) (#3078)"


sdottaka:
Thank you!

Issue #3090, opened 2025-12-23
clarification needed: when choosing 2 files/folders to compare from context menu, which go to left and right sides? how about when comparing 3?


tsehygit:
steps:
- highlight/choose 2~3 items in file explorer
- rightclick 1 of them
- click (WM icon) compare (C)

results: one of them appear in left panel and the other in right panel
question for this procedure: which go to left and right sides? how about when comparing 3?

sdottaka:
When multiple files are selected in Explorer, the file you right-click becomes the first (primary) item.
The remaining items follow in the order they were selected.
So, to make a file appear on the left side, select multiple files and right-click the one you want on the left.

See also: https://devblogs.microsoft.com/oldnewthing/20190409-00/?p=102406

sdottaka:
This behavior is now explained in the FAQ, so I am closing this issue.

PR #3094, opened 2025-12-24
Update English.pot RC to left side #3090

tsehygit:
thanks for the info. i think context menu should have this.
Closes: #3090


sdottaka:
Thank you for the PR. However, I’m afraid we can’t accept it at this time for the following reasons.
We’re considering adding an explanation to the manual FAQ instead.

- .pot files are generated automatically from Merge.rc and ShellExtension.rc, so changes should be made to those source files rather than editing the .pot file directly.
- Adding long explanatory text to menu labels is not ideal from a UI/UX perspective.

PR #3121, opened 2026-01-03
Update Brazilian.po

Marcellomco:
Added translation for "Add folder statistics filter attributes and helper menu (#3111)"


sdottaka:
Thank you!

PR #3130, opened 2026-01-10
Fixed Ukrainian translation. Changed several buttons to fit translated 'defaults' text.

YevgenBorodkin:
- Translated several hundreds of untranslated strings.
- Updated some of the existing translations to have them consistent.
- Changed buttons to fit the translated text: 'defaults' (Plugins settings, Columns settings, Patch generator settings). The "options" menu contains buttons with 'defaults' text that have enough room for the translated text.


sdottaka:
Thank you!

PR #3137, opened 2026-01-15
Update French translations in French.po

slolo2000:
Updated French translations for various UI elements in WinMerge.


sdottaka:
Thank you!

PR #3165, opened 2026-01-30
Fix French.po

gbabin:
"Copy All to Right" and "Copy All to Left" have the same French translation.

This PR fixes the translation of "Copy All to Left".


sdottaka:
Thank you!

PR #3171, opened 2026-02-03
Update Brazilian.po

Marcellomco:
Added translation for "Add automatic crash detection and reporting (#3167)"


sdottaka:
Thank you!

PR #3194, opened 2026-02-11
Update Brazilian.po

Marcellomco:
Added translation for "Add PrettifyPO and ValidatePO plugins for .po/.pot files"


sdottaka:
Thank you!

PR #3219, opened 2026-02-24
Update Brazilian.po

Marcellomco:
Added translation for "Add Replace Lists menu for Rename/Move detection keys (#3215)"


sdottaka:
Thank you!

Issue #3212, opened 2026-02-19
Feature request: In CompareMSExcelFiles do not append a sheet order number to individual tsv, txt or html  files


JoHoGH:
When using the option "Extract Workbook data to multiple files" , a number is appended to the sheet name, reflecting the order of the sheet in the workbook, e.g "(04)Sheet3.tsv".
 This causes a problem if one of the workbooks being compared has had a sheet or sheets inserted or deleted, or if sheets are moved, because the sheet tsv, txt or html filename is no longer the same. The result is that the sheets are not compared.
I'd prefer to have the "(" + No  + ")" removed from the file name. Or at least have the option to remove it.

sdottaka:
This issue has been resolved in commit c17d7ba by adding an option to include or omit the sheet order number in filenames.

PR #3231, opened 2026-02-26
Update Brazilian.po

Marcellomco:
Added translation for "Make sheet order number optional in extracted Excel filenames (#3212)"


sdottaka:
Thank you!

PR #3232, opened 2026-02-26
Update German.po

Havoc7891:
Updates the German translation for the string introduced in c17d7ba7 ("Make sheet order number optional in extracted Excel filenames")


sdottaka:
Thank you!

PR #3242, opened 2026-03-09
Add CFR Java decompiler plugin for .class file comparison

Gaurav-pasi:
Adds a new DecompileJVM unpacker plugin that uses CFR (Class File Reader) to decompile .class files into readable Java
   source code, enabling meaningful line-by-line comparison of compiled Java classes.

  This complements the existing DisassembleJVM plugin (javap bytecode output) by providing human-readable Java source
  output instead of raw bytecode instructions.

  The plugin appears under Plugins > Unpacker > Decompilation > "Decompile Java (CFR)" and outputs .java files for
  comparison. It also works with .class files inside JAR archives.

  Changes:
  - Plugins/Commands/CFR/cfr.bat - Launcher script with auto-download support for CFR 0.152
  - Plugins/Commands/CFR/URL.txt - Source URLs
  - Plugins/Plugins.xml - New DecompileJVM plugin entry
  - Docs/Manual/English/Plugins.xml - Plugin documentation

### DIFF
--- a/Docs/Manual/English/Plugins.xml
+++ b/Docs/Manual/English/Plugins.xml
@@ -1477,6 +1477,59 @@
         </segmentedlist></para>
     </section>
 
+    <section id="Plugins_DecompileJVM">
+      <title><filename>DecompileJVM<indexterm>
+          <primary>DecompileJVM plugin file</primary>
+        </indexterm></filename></title>
+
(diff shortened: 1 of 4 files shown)


sdottaka:
Thank you for the PR.
I will merge it for now.
One thing I noticed is that the downloaded file is not verified against the SHA256 value, so I may add a checksum verification later.

PR #3243, opened 2026-03-10
Add SHA256 checksum verification for CFR JAR download

Gaurav-pasi:
## Summary
- Add SHA256 checksum verification for downloaded CFR JAR using CertUtil, consistent with other plugin download scripts (yq, PlantUML, java, etc.)
- Delete the downloaded file on checksum mismatch to prevent executing unverified JARs
- Fix the incorrect SHA256 hash value to match the actual cfr-0.152.jar release
- Fix minor grammar in the DecompileJVM plugin documentation

## Context
Follow-up to #3242 -- addresses the review feedback from @sdottaka about missing SHA256 verification for the downloaded CFR JAR file.

## Test plan
- Verified correct SHA256 hash of cfr-0.152.jar downloaded from GitHub releases matches the updated value
- Verified CertUtil hash check returns True for the real JAR
- Verified CertUtil hash check returns False for a tampered/fake file
- Verified implementation pattern matches existing scripts (yq.bat, q.bat, PlantUML.bat, etc.)

### DIFF
--- a/Docs/Manual/English/Plugins.xml
+++ b/Docs/Manual/English/Plugins.xml
@@ -1482,7 +1482,7 @@
           <primary>DecompileJVM plugin file</primary>
         </indexterm></filename></title>
 
-      <para>Java decompiler with CFR - decompiles .class files to readable Java source code</para>
+      <para>Java decompiler with CFR - decompiles .class files into readable Java source code</para>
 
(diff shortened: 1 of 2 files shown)


sdottaka:
Thank you!

PR #3280, opened 2026-03-26
Add tree-sitter based syntax highlighting support

Thorium:
Current "highlight a keyword" is not enough for modern programming languages.
This PR integrates a tree-sitter as an optional syntax highlighting engine that supplements the existing keyword-based CrystalEdit parsers. When a grammar DLL and highlight query (.scm) are present in the TreeSitterGrammars directory, tree-sitter provides full AST-based highlighting; otherwise the existing parser runs unchanged.

Core components:
- TreeSitterParser.h/.cpp: CTreeSitterParser, CTreeSitterColorMap, CTreeSitterLanguage, and TreeSitterRegistry classes
- ParseLine virtual override in CMergeEditView for tree-sitter results
- Incremental parsing via ts_tree_edit() on each edit operation
- Lazy reparse with dirty flag (fires once per paint cycle)
- Status bar indicator showing [TS:language] in encoding pane
- Post-build step to copy grammar DLLs from Release to Debug/Test

Supported languages: bash, c, c-sharp, cpp, css, dtd, flow, fsharp, fsharp_signature, go, html, java, javascript, json, php, php_only, python, ruby, rust, tsx, typescript, xml. Many more can be easily added: https://tree-sitter.github.io/tree-sitter/

Grammar DLLs are built separately via build-grammars.ps1.

### Summary
1. When a file is opened, InitializeTreeSitter checks the file extension against TreeSitterRegistry's extension-to-language map
2. If a matching grammar DLL + highlight query exists in TreeSitterGrammars\, tree-sitter takes over syntax highlighting for that file
3. If no grammar is available, the existing CrystalEdit keyword-based parser (CCrystalTextView::ParseLine) runs exactly as before — zero change in behavior
So the 22 languages with grammar DLLs get tree-sitter highlighting, and all other languages (COBOL, Fortran, Pascal, etc.) keep using the existing keyword parsers. It's fully backward-compatible.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "Externals/darkmodelib"]
 	path = Externals/darkmodelib
 	url = https://github.com/winmerge/darkmodelib.git
+[submodule "Externals/tree-sitter"]
+	path = Externals/tree-sitter
+	url = https://github.com/tree-sitter/tree-sitter.git
--- a/Externals/crystaledit/editlib/TreeSitterParser.cpp
+++ b/Externals/crystaledit/editlib/TreeSitterParser.cpp
@@ -1,0 +1,1034 @@
+////////////////////////////////////////////////////////////////////////////
+//  File:       TreeSitterParser.cpp
+//  Version:    1.0.1
+//  Created:    2026-03-26
+//
+//  Tree-sitter based syntax highlighting bridge for CrystalEdit.
+//
+//  SPDX-License-Identifier: GPL-2.0-or-later
+////////////////////////////////////////////////////////////////////////////
+
+#include "StdAfx.h"
+#include "TreeSitterParser.h"
+#include "ccrystaltextview.h"
+#include "ccrystaltextbuffer.h"
+
+#include <tree_sitter/api.h>
+
+#include <fstream>
+#include <sstream>
+#include <cassert>
+
+#ifdef _DEBUG
+#define new DEBUG_NEW
+#endif
+
+// ============================================================================
+// CTreeSitterColorMap
+// ============================================================================
+
+CTreeSitterColorMap::CTreeSitterColorMap()
+{
+    // Map standard tree-sitter highlight capture names to WinMerge COLORINDEX.
+    //
+    // Tree-sitter highlight queries use capture names like @keyword, @string,
+    // @comment, etc. These names come from nvim-treesitter conventions.
+    // We collapse them into WinMerge's available color categories.
+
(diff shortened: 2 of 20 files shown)


sdottaka:
Thank you for the PR. It seems the following file is missing—could you please add it?
winmerge\Externals\tree-sitter\tree-sitter.vcxitems

Thorium:
The pattern seems to be putting these vcxitems `Externals` as submodules. 
Tree-sitter is a submodule in the WinMerge repo that points to the upstream https://github.com/tree-sitter/tree-sitter.git (the core tree-sitter library), pinned at tag v0.25.10.

I'm not very good at setting up submodules. So here it is:
[tree-sitter.zip](https://github.com/user-attachments/files/26289294/tree-sitter.zip)

sdottaka:
Thank you for providing the files. I couldn’t find build-grammars.ps1—could you please provide it as well?
I’m planning to merge this for now, but I may create a separate branch shortly afterward to continue integration work.

Thorium:
Added the build-grammars.ps1.
Also there is some well-known grammars (grammars.json) and one custom for testing (fsharp-highlights.scm).

sdottaka:
I created a feature/tree-sitter branch and merged this PR into it.

After testing, it seems that syntax highlighting for C++ files is not working correctly (see the attached screenshot). 

It may take some time before I can fully work on tree-sitter, so I would appreciate it if you could take a look.

image

Thorium:
Closing this in favour of the feature/tree-sitter branch, and added a new PR to fix the issue to target that.

Thorium:
Ok, here is a new version of build-grammars.ps1 that makes it cached and parallel, so happens in seconds, rather than hours: https://raw.githubusercontent.com/Thorium/winmerge/refs/heads/feature/tree-sitter-semantic-merge-v2/Tools/TreeSitterGrammars/build-grammars.ps1

sdottaka:
@Thorium 
Thanks for the suggestion and the script.

build-grammars.ps1 has now been moved into the WinMerge/tree-sitter-grammars repository, and is currently undergoing major refactoring and restructuring.

Because of that, I’m actively working on changes in a different direction at the moment, but I appreciate the idea and the effort.

Thorium:
That's fine, just wanted to share so that if you are still using the old, slow script.

For bigger "merge-conflicts", instead of a traditional merge, I think AI agents do quite a decent job of comparing the current branch with another one and spotting which changes would be beneficial to port.

PR #3283, opened 2026-03-28
Feature/tree sitter

Thorium:
Delegated an agent to check. The issue with C++ (only) was basically that it was needing both C grammar and C++ grammar, and the Powershell-script to build the grammars picked one, not both.

I tested, seems to work better now.


sdottaka:
Thank you!

PR #3327, opened 2026-04-19
Code review fixes for 5 oldest source files

JiayangLai:
## Summary

This PR contains code review fixes for 5 of the oldest source files in the codebase:

### Files Modified

1. **Src/Common/UnicodeString.cpp** - Fixed `strip_hot_key()` function
   - Bug: The function was erasing only 2 characters ("(&") but should erase 3 characters ("(&X") to fully remove the hotkey marker including the accelerator character
   - Change: `str2.erase(it, it + 2)` → `str2.erase(it, it + 3)`

2. **Src/Common/VersionInfo.cpp** - Fixed `GetVersionInfo()` function
   - Bug: `m_bVersionFound` was set to `true` before calling `GetFileVersionInfo()`, which could leave stale state if the function fails
   - Change: Moved `m_bVersionFound = true` inside the success block after `GetFileVersionInfo()` succeeds

3. **Src/Common/Shell.cpp** - Fixed `OpenParentFolder()` function
   - Bug: Filenames containing spaces were not properly quoted when passed to explorer.exe
   - Change: Added check for spaces and extra quoting when needed

4. **Src/Common/paths.cpp** - Fixed `GetPairComparability()` function
   - Bug: The fallback `DoesPathExist()` calls weren't passing the `IsArchiveFile` callback, causing inconsistent behavior
   - Change: Added `IsArchiveFile` parameter to all fallback `DoesPathExist()` calls

5. **Src/Common/Shell.h** - Reviewed, no issues found (header-only declarations)

### Testing

These changes fix bugs that could cause:
- Incorrect hotkey stripping in menu items
- False positive version info detection
- Explorer failure when opening parent folders of files with spaces in names
- Incorrect path comparison results when archive files are involved

All changes are backward compatible and don't alter the public API.

---
Generated by Claude Code

### DIFF
--- a/Src/Common/Shell.cpp
+++ b/Src/Common/Shell.cpp
@@ -46,6 +46,9 @@ void OpenFileOrUrl(const tchar_t * szFile, const tchar_t * szUrl)
 void OpenParentFolder(const tchar_t * szFile)
 {
 	String param = _T("/select,\"") + String(szFile) + _T("\"");
+	// Add extra quotes if filename contains spaces
+	if (param.find(_T(' ')) != String::npos)
+		param = _T("\"") + param + _T("\"");
 	ShellExecute(nullptr, _T("open"), _T("explorer.exe"), param.c_str(), nullptr, SW_SHOWNORMAL);
 }
 
--- a/Src/Common/UnicodeString.cpp
+++ b/Src/Common/UnicodeString.cpp
@@ -53,7 +53,7 @@ String strip_hot_key(const String& str)
 	String str2 = str;
 	auto it = str2.find(_T("(&"));
 	if (it != String::npos)
-		str2.erase(it, it + 2);
+		str2.erase(it, it + 3);
 	strutils::replace(str2, _T("&"), _T(""));
 	return str2;
 }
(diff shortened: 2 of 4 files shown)


sdottaka:
Thank you for the PR and for reviewing several older parts of the codebase — much appreciated.

However, after reviewing the changes, we cannot accept this PR as-is due to several issues:

- UnicodeString.cpp (strip_hot_key)
  The change misuses std::string::erase. The intention of this function is to remove everything from "(&" to the end of the string, so the correct fix would be `erase(it)`. The current change introduces a bug.

- Shell.cpp (OpenParentFolder)
  The added quoting breaks the command line passed to explorer.exe. The path is already correctly quoted, and adding outer quotes can prevent proper parsing.

- VersionInfo.cpp
  Moving `m_bVersionFound = true` is correct, but the variable should also be explicitly initialized/reset (e.g., set to false at the beginning of the function) to avoid stale state.

- paths.cpp (GetPairComparability)
  The change removes the fallback behavior that allows treating paths as normal files when archive detection is inconsistent. This is intentional behavior and important for usability, so this change introduces a regression.

While the intention to improve correctness and consistency is appreciated, some of these changes alter existing behavior or introduce new issues.

Please feel free to submit a revised PR addressing the above points.

PR #3337, opened 2026-04-23
Stabilize tree-sitter highlight precedence

Thorium:
## Summary
- make tree-sitter locals handling accept `local.scope.*`, `local.definition.*`, and `local.reference.*` capture names so newer query conventions still resolve symbols correctly
- add deterministic precedence metadata for overlapping captures so identical code is colored consistently across panes and languages instead of depending on query match order
- apply the same precedence handling to injected-language highlights so embedded regions follow the same stable coloring rules

## Validation
- reviewed the isolated diff against `origin/feature/tree-sitter`
- did not run a full WinMerge build in this clean worktree


sdottaka:
Thanks for the PR — I’ll merge it.

Merging into the Tree-sitter master branch may take some time. I’d like to address the following items first:

- The syntax highlighting with Tree-sitter currently looks quite different from CrystalEdit’s C++ highlighting; I’d like to resolve this.
- Using Tree-sitter to ignore comment differences in folder comparisons when it is enabled
- Integration into the installer
- Adding an option to enable/disable Tree-sitter
- Implementing “Go to definition” (F12) using Tree-sitter

Thorium:
I'd love to have semantic merge option/suggestions, 
So, something like this, file A:
```
f() { x + 1 }
g() { y - 2 }
```
B) Changes to
```
g() { y - 2 }
f() { x + 1 }
```
C) And changes to
```
f() { x + 3 }
g() { y - 2 }
```
...then we could auto-resolve to
```
g() { y - 2 }
f() { x + 3 }
```
...but the world may not be ready for that yet.

Or even the classic case where a file gets new additions to the same line numbers in B and C, and then the editor tries to cascade some weird BCBCBC instead of working BBBCCC or CCCBBB.

sdottaka:
I’ve been thinking it would be nice if we could:

Compare functions independently (e.g., sort them while preserving original line positions, then write back respecting those positions)
Extract functions into a separate view for comparison/editing and apply the changes back

Tree-sitter might make this kind of approach more feasible.

PR #3338, opened 2026-04-23
Tree-sitter runtime integration for compare views

Thorium:
## Summary
- wire tree-sitter runtime assets into Release builds and installer packaging so compare-view features are available from built binaries
- add compare-view tree-sitter UI integration, including the editor option, menu wiring, and position-aware Go to Definition enablement
- extend same-file navigation with tags query support

## Testing
- build `WinMerge.sln /t:Merge /p:Configuration=Release /p:Platform=x64`
- rebuild `Tools/TreeSitterGrammars` output and verify `TreeSitterGrammars` is present beside `WinMergeU.exe`


Thorium:
This PR addresses the earlier comment:

> * Using Tree-sitter to ignore comment differences in folder comparisons when it is enabled
> * Integration into the installer
> * Adding an option to enable/disable Tree-sitter
> * Implementing “Go to definition” (F12) using Tree-sitter

sdottaka:
Thanks for the quick implementation!

That said, it may take some time for validation and refactoring, so I expect merging into master will likely happen after the next stable release.

sdottaka:
Thanks for the updates to the PR — I’ll go ahead and merge it for now.

One note: the changes in Externals/crystaledit/editlib/TreeSitterWrapper.h introduce a dependency on UnicodeString.h, which is part of WinMerge. Since TreeSitterWrapper.h belongs to the external crystaledit module, it shouldn’t depend on WinMerge sources. I’ll fix this later.

PR #3340, opened 2026-04-25
Update German.po

Havoc7891:
Updates the German translation for the strings introduced in https://github.com/WinMerge/winmerge/commit/79168f8927f6af97fba0c004cd5664ab6a896172 ("Add recent items and clipboard history to header bar menu (https://github.com/WinMerge/winmerge/pull/3330)")


sdottaka:
Thank you!

PR #3350, opened 2026-04-29
Fix tree-sitter go to definition from context menus

Thorium:
## Summary
- move the caret to the symbol under the mouse before running context-menu navigation commands
- prefer the tagged definition target when a right-clicked type reference resolves to the current line instead of the earlier type definition
- fix right-click go-to-definition for type annotations


sdottaka:
@Thorium 

Could you please take a look at the Copilot review and fix any issues if needed?

sdottaka:
Thank you!

PR #3362, opened 2026-05-03
Update German.po

Havoc7891:
Updates the German translation for the string introduced in https://github.com/WinMerge/winmerge/commit/767ffa7903a662953e1f03c32fe2a16727eed949 ("Add clipboard URL handler and clipboard history menu (https://github.com/WinMerge/winmerge/pull/3352)")


sdottaka:
Thank you!

PR #3374, opened 2026-05-12
Add display line filter bar for file compare window

sdottaka:
# Add Display Line Filter Feature

Adds a display line filter bar to the File Compare window, allowing users to control which lines are visible without affecting the comparison result.

### New Features
- Display line filter bar (Ctrl+Shift+L) for File Compare window
- Filter expressions support line attributes: `Line`, `LineLength`, `LineNumber`, `LineFlags`, `EOL`, `EOLStr`, etc.
- Context-aware functions: `matchContext()`, `matchNumber()`, `matchBlockNumber()`, `matchBlockCount()`, `matchInside()`, etc.
- Right-click "Add to Display Filters" to build filter expressions from selected text
- Display folder filter for Folder Compare window

### Improvements to Filter Expression Engine
- Renamed `EvalContext` to `FilterEvalContext`, adding `ILineDataProvider` and `FilterSharedContext` for line-level evaluation
- Added `none` literal
- Added `trim()`, `trimLeft()`, `trimRight()`,`strCount()`,`strFind()`,`regexCount()`,`regexFind()`,`regexExtract()`  string functions
- Added `EOL`/`EOLStr` attributes for file and line comparison
- Added `ParseDateTime()` supporting additional date formats (`YYYY-MM-DD` etc.)
- Refactored function documentation into categorized sections

### Code Refactoring
- Extracted `CFilterHelperMenuBase` as a base class shared by `CFileFilterHelperMenu` and `CLineFilterHelperMenu`
- Extracted common menu utilities into `FilterMenuHelpers` namespace
- Extracted `LineFilterHelper`, `LineFilterBar`, `ReplaceListHelper` as separate modules

## Notes
- The display filter is applied after the diff context filter
- `matchInside()` evaluation is line-based; multiple `/*` or `*/` occurrences within a single line are not distinguished

## Examples

Expression | Description
-- | --
ERROR | Show only lines containing "ERROR"
le:Different | Show only lines with differences
le:Different and LeftMissing | Show only lines where right side is added (left side is deleted)
le:Different and Line contains "ERROR" | Show only lines with differences containing "ERROR"
le:matchContext(Different and Line contains "ERROR", 3, 3) | Same as above, with 3 lines of context above and below
le:matchInside(Line contains "BEGIN", Line contains "END") | Show only lines within BEGIN...END log blocks
le:LineLength > 80 | Show only lines exceeding 80 characters
le:EOLStr = "CRLF" | Show only lines with Windows (CRLF) line endings
[...]

PR #3387, opened 2026-05-25
Update German.po

Havoc7891:
Updates the German translation for the strings introduced in bf84405 ("Add display line filter bar for file compare window (#3374)")


sdottaka:
Thank you!

PR #3409, opened 2026-06-07
Update Italian.isl

Need74:
Just because I know a little Italian and I just uploaded the French file, which was missing.

I obviously got the file from https://jrsoftware.org/files/istrans/ Best regards

### Summary
Briefly describe what this PR changes.

(Optional) Fixes #<issue-number>


sdottaka:
Thank you for the PR.

The updated Translations/InnoSetup/Unbundled.is5/Italian.isl appears to be based on the Inno Setup 6.5 translation file. Since Unbundled.is5 is intended for Inno Setup 5, I don't think the 6.5 version should be applied here.

Could you confirm the source of this translation file?

Need74:
> Thank you for the PR.
> 
> The updated Translations/InnoSetup/Unbundled.is5/Italian.isl appears to be based on the Inno Setup 6.5 translation file. Since Unbundled.is5 is intended for Inno Setup 5, I don't think the 6.5 version should be applied here.
> 
> Could you confirm the source of this translation file?

Hello, I realize that in my haste I made a mistake and that this PR is irrelevant since the version is incorrect.

I will try sending a message directly to <bovirus@gmail.com> so that he can handle this himself if the need exists for the Italian version.

Sincerely, and sorry for my mistake.

PR #3413, opened 2026-06-12
Tree-sitter: correctness fixes, option gating, and editor UX features

Thorium:
Bug fixes

- Comment filter ("Ignore comment differences"): parse contexts are now created lazily inside `CDiffWrapper::PostFilter` from the exact text being diffed, fixing three bugs at once: 3-way compares queried the wrong file's parse tree (pair scripts are (1,0)/(1,2)/(0,2), contexts were always [0]/[1]), merge-editor rescans used ghost-line-shifted line numbers, and editor rescans looked up the grammar from the .tmp temp-file name (grammar selection now uses the original compare paths via `SetCompareFiles`). Decodes with `convertToTString` so positions match the queried text.
- Lifetime/memory: dynamic TextDefinitions are freed with the matching deleter (name/exts leaked before) and views are re-pointed before a definition is freed (use-after-free on reload to a grammar-less extension).
- Incremental parsing safety: an edit arriving while the parse cache is already dirty discards the tree and forces a full reparse instead of applying `ts_tree_edit()` against stale byte offsets; undo/redo handlers discard the tree explicitly via `DiscardTree()`.
- Thread safety: grammar registry is mutex-guarded (folder-compare workers load grammars concurrently); extension lookup is case-insensitive.
- Option gating: `OPT_TREE_SITTER` now flows through `DIFFOPTIONS::bTreeSitterCommentFilter`, so disabling tree-sitter consistently affects the editor, folder compare, and patch generation.

New features

- Status bar breadcrumb: shows the active grammar and enclosing symbol at the caret, e.g. `[TS:cpp ▸ MyClass::Foo]` (AST walk; supports name and nested declarator fields, requires a body to exclude forward declarations/invocations).
- Symbol occurrence highlighting: all occurrences of the identifier under the caret are highlighted in every pane via a transient marker (AST-confirmed identifiers only — not comments/strings/keywords; never persisted or shown in the marker dialog). Uses a new lightweight `CCrystalTextMarkers::InvalidateViews()` repaint instead of the cache-resetting `UpdateViews()`.

---

**2026-06-12: rebased onto current `feature/tree-sitter`.** Parts of the original commit that the base branch has since implemented independently were dropped: the UndoRecord decoupling (superseded by `ITextBuffer`/`TextEdit`, commit 68b5e200d), the PostFilter cookie-ordering restore (5096704a6), the `m_pBuffer` initialization (member removed by the stateless refactor, 53b3e0a1e), and the undo-position synchronization (`AddUndoRecord` now notifies the parser of every record, so the dirty-state check above covers multi-edit staleness). The accidental `Externals/jq` pointer bump and `Externals/tree-sitter` gitlink are also removed.

### DIFF
--- a/Externals/crystaledit/editlib/ccrystaltextbuffer.h
+++ b/Externals/crystaledit/editlib/ccrystaltextbuffer.h
@@ -156,7 +156,7 @@ public :
     };
     std::shared_ptr<SharedTableProperties> m_pSharedTableProps;
 
-	void* pParseContext; // context for incremental parsing, owned by the parser
+	void* pParseContext = nullptr; // context for incremental parsing, owned by the parser
 
     //  Helper methods
     void InsertLine (const tchar_t* pszLine, size_t nLength, int nPosition = -1, int nCount = 1);
--- a/Externals/crystaledit/editlib/ccrystaltextmarkers.cpp
+++ b/Externals/crystaledit/editlib/ccrystaltextmarkers.cpp
@@ -74,6 +74,19 @@ void CCrystalTextMarkers::UpdateViews()
 		pView->UpdateView(nullptr, nullptr, 0, -1);
 }
 
+/**
+ * @brief Repaint all attached views without resetting their caches.
+ *
+ * Markers only affect colors, not parse state, line lengths or wrapping,
+ * so a plain repaint is enough to show a marker change. Use this instead
+ * of UpdateViews() for frequent changes (e.g. caret-following markers).
+ */
+void CCrystalTextMarkers::InvalidateViews()
+{
+	for (auto& pView : m_views)
+		pView->Invalidate(false);
+}
+
 CString CCrystalTextMarkers::Serialize() const
 {
 	CString text;
--- a/Externals/crystaledit/editlib/ccrystaltextmarkers.h
+++ b/Externals/crystaledit/editlib/ccrystaltextmarkers.h
@@ -36,6 +36,7 @@ public:
 	void AddView(CCrystalTextView *pView);
 	void DeleteView(CCrystalTextView *pView);
 	void UpdateViews();
+	void InvalidateViews();
 	void SetEnabled(bool enabled) { m_enabled = enabled; };
 	bool GetEnabled() const { return m_enabled; };
 	CString MakeNewId() const;
--- a/Externals/crystaledit/editlib/ccrystaltextview.h
+++ b/Externals/crystaledit/editlib/ccrystaltextview.h
@@ -769,6 +769,7 @@ public :
 
     //  Source type
     CrystalLineParser::TextDefinition *m_CurSourceDef;
+    CrystalLineParser::TextDefinition *GetTextType () const { return m_CurSourceDef; }
     virtual bool DoSetTextType (CrystalLineParser::TextDefinition *def);
     virtual bool SetTextType (const tchar_t* pszExt);
     virtual bool SetTextType (CrystalLineParser::TextType enuType);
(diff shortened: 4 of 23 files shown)


sdottaka:
Thank you for the PR.

At the moment I'm working on PR #3398, which is intended to synchronize the Tree-sitter branch with the current master branch. The branch has diverged quite a bit from master, and resolving the remaining issues is taking longer than expected. I currently estimate that the work will take at least a few more weeks.

Your PR contains substantial changes to the Tree-sitter implementation, and many of those changes are likely to conflict with the ongoing work in #3398. Because of that, it would be difficult for me to review and merge this PR in its current form before the master synchronization is completed.

Would it be possible to close this PR for now and reapply/rebase the changes after #3398 has been merged? I apologize for the inconvenience, but that would make integration much easier.

Also, regarding some of the new features:

* Due to current performance concerns, Tree-sitter will no longer be the preferred parser by default when a built-in parser is available. Tree-sitter will still be used for languages that do not have a built-in parser. If we later introduce asynchronous background parsing, we may be able to revisit this decision.
* I have also removed the status bar language display for now. My current thinking is that, if we expose the selected language in the status bar, it should eventually be clickable so users can directly change the language from there.

Thank you again for your work and contributions.

Thorium:
Sounds good, let's get some tree-sitter to main, and then we can improve it later.

Thorium:
I've taken this in my experimental branch https://github.com/Thorium/winmerge/tree/feature/tree-sitter-semantic-merge-v2 way further (and it actually works).
The user can use semantic merges (automatic if toggled on, or manually via the menu), in code files. 

So, for example, there is a method that is moved (BuildMessage), and now the other version changes the content, you still can synchronise contexts:

image

Now, the option gives you a dialog of changes (ok, the UI is not yet perfect), which asks you to accept the semantic merging:

image

So these kinds of ideas:
- How often do you try to merge 2 changed files, and the auto-merge "take all" is cutting the other function in half, instead of just putting it to "after" or "before" the other one. If you know where a function starts and ends, then no need to cut-half.
- What if you have a long living branch (anti-pattern), that gets changes that are still tried to sync to new branch that has moved on.
- ...and so on.

Thorium:
Closed this, because the only relevant change was fixed as a separate small PR.

My semantic merge branch is also now rebased to the latest `feature/tree-sitter`
https://github.com/Thorium/winmerge/tree/feature/tree-sitter-semantic-merge-v3
If you want, I can do PR about that later.

sdottaka:
@Thorium 

I merged the tree-sitter branch into master. Thank you for your PR.
The tree-sitter feature is planned for 2.16.57 Beta, scheduled for release on July 5, 2026.

Until the stable release 2.16.58 (planned for July 27, 2026), I intend to focus on other areas and hold off on tree-sitter feature additions beyond bug fixes and new language support. There was a crash I couldn't reproduce, so I'll be working on stabilization and reducing highlighting differences from CrystalEdit.

For future tree-sitter feature additions, I'm considering the following:
- Select and extract specific methods, functions, or classes for comparison
- Filter display lines by function or class name using display line filter expressions
- Extend Go to Definition to support files outside the currently diffed files

Regarding https://github.com/Thorium/winmerge/tree/feature/tree-sitter-semantic-merge-v3 -  tried it briefly but couldn't get it working as the menu items were disabled. Interesting ideas, though.

If you do create a PR, it looks like this branch bundles several features, so it would be helpful if you could split it into separate PRs. As you may know from the tree-sitter branch, large PRs take considerable time and effort to integrate. Please also note that we may decide not to merge a PR even after review — for example, we're uncertain about Ignore Trailing Punctuation and the status bar changes.

PR #3427, opened 2026-06-22
Tree-sitter: fix data race in TreeSitterRegistry (folder-compare workers)

Thorium:
## Problem

`TreeSitterRegistry` (a singleton) loads grammars lazily: `Initialize()` populates `m_availableLanguages`, and `GetLanguageForName()` lazily loads grammar DLLs, mutating `m_languages` / `m_failedLanguages` / `m_bInitialized` on cache misses. None of this is synchronized, yet it is reached concurrently from folder-compare worker threads via `CDiffWrapper::PostFilter` (the ThreadPool runs one `FolderCmp` per worker). The first concurrent use of a not-yet-loaded grammar races on the shared maps/set — a data race that can corrupt the containers or crash during folder compares of source trees.

## Fix

Guard the registry's lazy state with a `std::mutex`:

- add `std::mutex m_mutex` to `TreeSitterRegistry`
- take a `std::lock_guard` at the top of `Initialize()` and `GetLanguageForName()`

No deadlock risk: `CTreeSitterLanguage::Load()` / `GetHighlightQuery()` do not call back into the registry, and no registry method calls another locked registry method, so the non-recursive mutex is safe.

This is the one still-applicable item from an audit of the older `feature/tree-sitter-fixes` work (PR #3413) against the current `feature/tree-sitter` — the registry race was never addressed by the intervening refactors. The other items from that branch are already superseded by the `LangServices`/`ISyntaxParser` architecture and `OPT_TREE_SITTER_MODE` gating.

## Testing

Builds clean: MSBuild `WinMerge.sln -t:Merge -p:Configuration=Release -p:Platform=x64` → 0 errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Src/TreeSitterParser.cpp
+++ b/Src/TreeSitterParser.cpp
@@ -1444,6 +1444,8 @@ TreeSitterRegistry& TreeSitterRegistry::Instance()
 
 void TreeSitterRegistry::Initialize(const std::wstring& sGrammarDir)
 {
+	std::lock_guard<std::mutex> lock(m_mutex);
+
 	if (m_bInitialized)
 		return;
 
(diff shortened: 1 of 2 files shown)


Thorium:
Picked a relevant item from #3413

sdottaka:
Thanks for the PR. I've merged it.

I may replace `std::mutex` with the POCO C++ Libraries mutex later to reduce the executable size.

I'm still working on syntax highlighting and performance tuning in the tree-sitter branch, so it may take a little while before the branch is merged into `master`.

PR #3446, opened 2026-07-05
WinMerge - Doc - Italian update

bovirus:
@sdottaka 

### Summary

WinMerge Doc - Italian language update


bovirus:
Could you please create a list of 

- Tags that contain text that translator cannot change
- Tags that contain text that translator can change

Please add these info in the header of file.

sdottaka:
Thank you for the suggestion.

I'm not sure if it is necessary to add this to the header of every DocBook file. Instead, I think it would be better to have a general translation guideline.

For example, translators should not translate:

* Product names (e.g. WinMerge, Windows, Microsoft Excel)
* File names, folder names, file extensions, and URLs
* XML tag names and attribute names
* `id` and `linkend` attribute values
* Plugin names (e.g. `CompareMSExcelFiles`)
* Command-line option names (e.g. `/new`, `/self-compare`)
* File filter keywords (e.g. `f:`, `d:`, `pf:`) and display filter prefixes (e.g. `le:`)
* Filter function names (e.g. `replace`, `substr`, `toHalfWidth`) and attribute names (e.g. `Name`, `Size`, `Date`)
* Code fragments and API/function names (e.g. `strlen`)

For example:

```xml
<xref linkend="Plugins_available" />
<section id="Filters_MaskApply">
<trademark class="registered">Microsoft</trademark>
<application>Excel</application>
<function>strlen</function>
<literal>toHalfWidth(expression)</literal>
```

I think these guidelines would be more useful than listing specific XML tags.

PR #3480, opened 2026-07-21
Update German.po

Havoc7891:
Restore and update German translation strings.


sdottaka:
Sorry for breaking the translations, and thank you for updating the German translation!

PR #3499, opened 2026-07-31
Add MiniMax provider support to AI plugin

octo-patch:
Reason: The AI text conversion plugin cannot select MiniMax or route requests to MiniMax regional endpoints.

- Add MiniMax provider choices for the global and China endpoints.
- Add MiniMax-M3 as the default selectable model for both choices.
- Make API key labels provider-neutral while preserving existing settings.

Checks:
- `git diff --check origin/master...HEAD`
- Node-based AI.sct script, provider configuration, and request routing validation
- Scriptlet structure validation with `xmllint --noout -`
- `msgfmt --check --check-format -o /dev/null Translations/WinMerge/English.pot`

### DIFF
--- a/Plugins/Strings.rc
+++ b/Plugins/Strings.rc
@@ -199,7 +199,7 @@ BEGIN
                             "Windows Registry URL Scheme Handler.\r\nArguments: reg.exe command line options."
     IDS_PLUGIN_DESCRIPTION54 "Basic text functions"
     IDS_PLUGIN_DESCRIPTION55 "AI-assisted text conversion"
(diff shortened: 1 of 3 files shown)


sdottaka:
Thank you for the PR!

I noticed one potential issue. If the user switches the provider to MiniMax but forgets to change the API key environment variable name, the plugin will still use `OPENAI_API_KEY`, so the OpenAI API key may be sent to the MiniMax endpoint.

Could you update the default environment variable name (e.g. `MINIMAX_API_KEY`) when the provider is changed?

octo-patch:
Thanks for the review. I updated provider switching to synchronize the MiniMax API key with the loaded credential value, then pushed a new commit. The diff check passed.

sdottaka:
Thank you!

PR #3505, opened 2026-08-03
Update ChineseTraditional.po

love80312:
### Summary
*(Note: Translated into English by AI)*

* Corrected and unified localization strings, and translated empty/untranslated strings.

### Issue / Question

It looks like the label is wrapping to a new line and getting truncated due to insufficient container width.
(看起來像是因為區塊寬度不足，導致標籤文字自動換行並被截斷了。)

螢幕擷取畫面 2026-08-03 161544

螢幕擷取畫面 2026-08-03 163422


sdottaka:
Thank you for pointing that out. I've already increased the label width so the label should no longer wrap or get truncated in commit 3c357ab.

PR #3512, opened 2026-08-04
Merge-result panel with option to do KDiff3 style 3-way merge

### LINKED ISSUE #3507: [FEATURE] KDiff style 3-way merge?
### Description

Hi, I love WinMerge comparing local files, but for my Git commits I still prefer the gramma's old KDiff3 (from 2014!).

So, my question is, could we have an option to have as intuitive 3-way merge?

Image

It's basically, these 3 windows next to each other:
(A: original, the common-ground) (B: my-changes) (C: incoming-updates)
And bottom:
(what is the result of the merge)
Then you just navigate the next conflict and select A, B or/and C. (Typically B and C.)

### Examples or references (optional)

_No response_

### Additional context

_No response_

Thorium:
Initial implementation

3-way text compares get an editable Merge Result Pane at the bottom (kdiff3 layout: 3 read-only sources on top, merge output below). Non-conflicting changes auto-merge on open; conflicts get a <Merge Conflict> placeholder you resolve with Choose Left/Middle/Right (Ctrl+Shift+1/2/3), choose-all, or by editing the result directly. Save Merge Result writes to the -o / conflict-file path, so it works as a git mergetool out of the box.

image

How it works

The result is a 4th text buffer that is not a diff input — the existing 3-way diff engine is untouched. A segment table maps result lines ↔ diffs and tracks per-diff resolution state (auto / chosen / conflict / hand-edited), kept in sync through buffer edit hooks (undo/redo included).
UI follows the existing Diff Pane pattern: a dockable bar (View → Merge Result Pane, visibility persisted), one full-width editable view with segment-state highlighting, click-to-jump and current-diff scroll sync.
While the pane is visible the three sources are forced read-only (restored on hide); middle pane is treated as the common ancestor, matching existing Auto Merge semantics. Hidden pane = stock WinMerge behavior; 2-way compares unaffected.
Guardrails: unresolved-conflict warning on save, unsaved-result prompt on close, opening with an output path auto-shows the pane.
Scope/limitations: result pane not shown in Location pane; renders plain text in table/CSV mode; changing diff options after hand-editing the result disables Choose for stale links instead of discarding edits; new strings are English-only pending translation.

Fixes #3507


sdottaka:
Thanks for working on this. It's interesting to see a KDiff3-style merge implemented with relatively few changes.
I have a few concerns about the current design:

* I don't think the fourth pane should always be shown for every 3-way comparison. Sometimes users simply want to compare three files, or edit the existing three panes without performing a merge. It might be better to enable the fourth pane only when explicitly requested, for example via a command-line option or when the Auto Merge button is pressed. KDiff3 takes a similar approach by letting the user choose whether to perform a merge in the Open dialog, where there is a checkbox to enable merging and a field to specify the output file.
* The fourth pane currently has no context menu. My first instinct was to right-click there to perform merge operations, but nothing appeared, which was a bit confusing.
* In practice, resolving merge conflicts rarely consists of simply choosing one side. Small manual edits are almost always required after selecting the base changes. When I tried it, I couldn't edit the fourth pane by typing, so I'd like the merge output to accept direct text editing, similar to KDiff3.
* The Merge menu has grown quite large. It may be worth showing the conflict-resolution commands only in the 4-pane mode, or grouping them into a submenu.
* We also need to consider the existing command-line options. Currently `/fl`, `/fm`, and `/fr` set the active input pane, while `/al`, `/am`, and `/ar` perform auto-merge into one of the existing panes. Introducing a dedicated output pane changes the semantics of these options, so I think the command-line behavior should be designed carefully as well.

Overall, I think the idea is promising, but I'd prefer to spend a bit more time refining the UI and behavior before merging it into the master branch.

Thorium:
Thanks for the feedback.
If we change the command line (or default options), we'll have a long tail of chasing different git-tool setups and instructions.

image

sdottaka:
I agree that we should avoid breaking existing git-tool setups and workflows.

In the screenshot you attached, the command line includes `-o "$MERGED"`. When `-o` is not specified, I don't think the fourth pane should be shown because the user is clearly just comparing three files. On the other hand, if `-o` is specified, we already know the user intends to perform a merge, so showing the fourth pane would make sense.

Ideally, when performing a merge, the focus would be on the fourth (output) pane. However, your screenshot also uses `-fm`, which currently means "focus the middle pane". Changing the meaning of existing options such as `-fl`, `-fm`, or `-fr` would be confusing and could break existing workflows. I'd prefer to keep their current behavior and, if necessary, introduce a new option (for example, `-fb`) to set the focus to the output pane.

sdottaka:
WinMerge also supports three-way comparison for binary comparison and image comparison, not only text comparison. Do you plan to support the fourth merge pane for these comparison modes as well? I think it would also be acceptable if the fourth pane is supported only for text comparison.

Thorium:
I think the text is enough for now.

sdottaka:
Thanks for the update.

For now, I'd like to merge this into the `feature/4pane-merge` branch instead of `master`.

Could you use this branch for a while and continue refining the feature based on your own experience? I think that will help identify usability issues before we consider merging it into `master`.

Here are some things I noticed while testing:

* After **Choose Left/Middle/Right**, performing **Undo** restores the text, but the conflict highlighting becomes incorrect. As shown in the attached screenshot, the red conflict background expands beyond the actual conflict, and ignored-difference colors sometimes appear unexpectedly.
* After Undo, enabling **Display Line Filters** (Ctrl+Shift+L) causes a crash.
* View options such as **View → View EOL** are not applied to the Merge Result pane.
* When the Merge Result pane has focus, pressing **Esc** does not close the window.
* The brown change marker in the left margin is shown for every line, making it difficult to identify the actual modified lines.
* Entering the 4-pane merge mode via **View → Merge Result Pane** feels a little difficult to discover. It might be more intuitive to provide a command in the **Merge** menu specifically for starting a merge session. Likewise, clicking **Auto Merge** could automatically switch to the 4-pane view.
* Auto Merge is currently performed even when no `/al`, `/am`, or `/ar` command-line option is specified. I think this changes the existing command-line behavior. When no auto-merge option is given, I would expect WinMerge to simply open the three input files for comparison. The merge result pane could still be shown if appropriate, but the merge itself should not be performed automatically.

image

Thorium:
Thanks, sounds good. I'll set this as my default merge tool for now, although I'm not currently working in an environment where there would be so many merges.

PR #3515, opened 2026-08-05
Merge-result panel with option to do KDiff3 style 3-way merge (2)

sdottaka:


Thorium:
I'm starting to be quite ok with the level of this feature, have nothing to edit right now.

PR #3517, opened 2026-08-07
Speed up diff view rendering for large contiguous hidden sections

KCMertens:
Diff rendering would process every line regardless of visibility. For some degenerate cases this could be hundreds of thousands or even millions of lines (sparse diffs in large files). Detect hidden lines and skip ahead.

Personally ran into this when trying to diff some ~100MB XML files with only about a dozen differences, scrolling across diff chunk boundaries tooks seconds. There's still some hidden work happening for e.g. syntax highlighting which needs to traverse backwards into hidden regions, but that seems inescapable.

### DIFF
--- a/Externals/crystaledit/editlib/ccrystaltextview.cpp
+++ b/Externals/crystaledit/editlib/ccrystaltextview.cpp
@@ -2704,6 +2704,23 @@ OnDraw (CDC * pdc)
   int nCurrentLine = m_nTopLine;
   while (rcLine.top < rcClient.bottom)
     {
+      // Invisible lines have no screen height.
+      // Directly skip to the next visible line instead of iterating perhaps millions of hidden lines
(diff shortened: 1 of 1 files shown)


sdottaka:
Thanks for the optimization and for the detailed explanation.

I profiled the rendering and confirmed that `RectVisible()` accounts for a significant portion of the time in this case, so avoiding unnecessary calls to it should provide a substantial improvement.

I'll merge this PR. I may make a small follow-up adjustment later to simplify or further optimize the handling of invisible lines.

aleksazr:
I see this is closed, but I'l ask here:
Will this speed up Ctrl+Shift+L (Display filter bar) results also? I have 2 files with 200k lines each, and I filter to see only lines containing "error" for example. If there are gaps, scrolling is slow.

sdottaka:
@aleksazr 
Yes, this optimization should also speed up scrolling when using the Display Filter Bar, especially when there are large gaps between visible lines.

PR #3522, opened 2026-08-08
Feature/4pane merge

Thorium:
### Summary

a) provenance markers to display what came from which file in the merge output window

b) when saving unresolved Merge Conflict, don't save it as `<Merge Conflict>` but instead use git/kdiff-style notation so data is not lost and you have the option to load and continue later. In the Merge Output Window, render it still as `<Merge Conflict>` but also provide user an option to "Show full conflict sections".


sdottaka:
Thank you!

PR #3524, opened 2026-08-09
Update German.po

Havoc7891:
Updates the German translation for the string introduced in https://github.com/WinMerge/winmerge/commit/0fb487d0d9e683afb583920e4771a94ce2da3312 ("Support adding plugins to pipelines from menus (https://github.com/WinMerge/winmerge/pull/3523)")


sdottaka:
Thank you!

PR #3541, opened 2026-08-12
Update German.po

Havoc7891:
Updates the German translation for the strings introduced in https://github.com/WinMerge/winmerge/commit/eaa8764dad6b26708f9239a4d750f05f61cb55dd ("Add an option to prefer the WIC decoder when loading images for image comparison. (#3537)") and https://github.com/WinMerge/winmerge/commit/19212898acb02301c435d326b4324843bdb82d01 ("Add MiniMax provider support to AI plugin (#3499)")


sdottaka:
Thank you!

PR #3576, opened 2026-08-25
Update German.po

Havoc7891:
Updates the German translation for the strings introduced in https://github.com/WinMerge/winmerge/commit/46c66d3575968be3a897601e2364424479c1abe7 ("Add column range filtering and ignore columns in comparison (https://github.com/WinMerge/winmerge/pull/3565)")


sdottaka:
Thank you!

Issue #3591, opened 2026-08-29
[FEATURE] Remember the image compare splitter position


Ikkoru:
## Description

The image compare splitter always lays out as an equal split. A dragged position is discarded by:

* resizing the window
* maximizing or restoring
* toggling **View > Split Vertically**
* opening the next image comparison
* restarting WinMerge

Expected: the position is kept.

Use case: **Dragging Mode > Vertical Wipe** with one pane reduced to a narrow strip (would also be nice if it could be made smaller than 32px).

Opened https://github.com/WinMerge/winimerge/pull/39
Have a WinMerge PR ready and waiting for your approval.

## Examples or references (optional)

WinIMerge, at `57ef929`:

| Location | Relevance |
| --- | --- |
| `ImgMergeWindow.hpp` `CalcChildImgWindowRect()` :1264 | Computes the equal split. Stateless |
| `OnSize()` :1406 | Calls it on every `WM_SIZE` |
| `MoveSplitter()` :1308 | Applies a drag to the child rects only |
| `SetHorizontalSplit()` :213 | Re-lays out through the same function |
| `WinIMergeLib.h` | No splitter property. `GetPaneWindowRect()` has no setter counterpart |

WinMerge:

| Location | Relevance |
| --- | --- |
| `Src/ImgMergeFrm.cpp` `LoadOptions()` | Runs in `OnCreateClient()`, before `OpenImages()` |
| `Src/ImgMergeFrm.cpp` `SaveOptions()` | Runs from `DestroyWindow()` |
| `Src/OptionsDef.h`, `Src/OptionsInit.cpp` | Existing `Settings/Image...` options |

No existing request found in either repository.

## Additional context

The value must live in `CImgMergeWindow` and be read by `CalcChildImgWindowRect()`. Storing it only on the WinMerge side does not survive, because `OnSize()` rebuilds the layout from scratch.

Two repositories: a WinIMerge change, then a submodule bump and the options wiring in WinMerge.

The interface is declared in three files. `Src/ImgMergeFrm.h` includes `"WinIMergeLib.h"`, which resolves to WinMerge's tracked copy at `Src/WinIMergeLib.h`, not the one in `Externals/winimerge`. Patching only the submodule copy builds WinIMergeLib and then fails in `Merge.vcxproj` with `C2039: 'SetSplitterPosition': is not a member of 'IImgMergeWindow'`.

### Design decisions

| Decision | Options |
| --- | --- |
[...]

PR #3593, opened 2026-08-30
Remember the image compare splitter position

Ikkoru:
Refs #3591. winmerge/winimerge#39 is merged; `Externals/winimerge` on master still points at 57ef929, so it needs a bump before this builds. No submodule change is included here.

Loads the splitter position in `CImgMergeFrame::LoadOptions()` and writes it back in `SaveOptions()`.

`Settings/ImageSplitterPosition` holds the smaller pane's size in pixels, signed to indicate which pane, so the range allows negative values.

`Src/WinIMergeLib.h` is the copy of the interface that `Src/ImgMergeFrm.h` includes, so the accessors are declared there as well as in `Externals/winimerge`.

Scoped to image compare, with no menu item. If it is more useful as a starting point for the wider Window > Keep Splitter Position design than as a merge, please treat it that way.


sdottaka:
Thank you!

Issue #3500, opened 2026-08-01
[FEATURE] Unpacker SelectLines support literal string


snowman:
### Description

By default, `SelectLines` use regular expression pattern, when comparing the Windows file path, I need to escape every `\` with `\\`

Add support for `-F` option like `grep` to enable literal match pattern

```bash
$ man grep | grep -A2 -- '-F, --fixed-strings'
     -F, --fixed-strings
            Interpret PATTERNS as fixed strings, not regular expressions.
```

snowman:
From the documentation https://manual.winmerge.org/en/Plugins.html#Plugins_Unpacker_SelectLines

`SelectLines [-v] [-i] [-e patterns]`
The `patterns` seems actually single `pattern`, to use multiple patterns, you need to quote them and join with `|` like `"PATTERN_1|PATTERN_2"`, notice the `|` must quote otherwise it'll interpret as Unpacker plugin command.

## Test
The file being test:

```sh
$ cat dc.txt
Line AAA
Line BBB
ignore me
Line CCC

$ cat dd.txt
ignore me
Line BBB
Line CCC
Line AAA
```
Previously I use space to separate patterns like `-e pattern_1 pattern_2` work but `-v -e pattern_1 pattern_2` doesn't work as expected

<img width="906" height="991" alt="Image" src="https://github.com/user-attachments/assets/ccf45859-f345-44e5-b15e-f450696bd386" />

snowman:
so literal string seems impossible, so close this issue

sdottaka:
Just for reference, multiple patterns can also be specified by repeating the -e option:

`SelectLines -e me -e AAA`

<img width="817" height="643" alt="Image" src="https://github.com/user-attachments/assets/e292c944-9150-43e7-a3de-c2700108f5b0" />

snowman:
okay, you can use multiple `-e pattern`, I didn't try it before!
Even if it doesn't support multiple `-e` option, you can still pipe with multiple `SelectLines` commands.
so the literal string option is possible 👍

sdottaka:
The `-F` option has been implemented in PR #3598, so I’ll close this issue. Thank you for the suggestion!

PR #3603, opened 2026-09-01
Update German.po

Havoc7891:
Updates the German translation for the strings introduced in https://github.com/WinMerge/winmerge/commit/6024cf235385d0dc2c944ff890e5f1eb2ad45856 ("SelectLines plugin: Add literal string matching with -F option. refs https://github.com/WinMerge/winmerge/issues/3500 (https://github.com/WinMerge/winmerge/pull/3598)")


sdottaka:
Thank you!

PR #3624, opened 2026-09-08
Add Azerbaijani translation

jamalkamaladdin:
This PR adds an Azerbaijani (az) translation.

New file `Translations/WinMerge/Azerbaijani.po`: 1821 entries, every msgstr filled in, msgid order identical to `Translations/WinMerge/English.pot`.

The header sets `Language: az` and `Plural-Forms: nplurals=2; plural=(n != 1);`.

Bracketed UI labels keep their brackets and translate the word inside, the same way `Turkish.po` and `German.po` do it, for example `<None>` becomes `<Heç biri>`.


sdottaka:
Thank you for the Azerbaijani translation!

Issue #2564, opened 2024-11-30
is there a way to Display File Names on Both Left and Right Sides in Folder Comparison


guitarbug:
Hi Team,
Thank you for creating such a fantastic tool! I have a suggestion to improve the Folder Comparison feature.

Currently, Filename are displayed in a **single column**. It would be incredibly helpful if Filename could be shown in **both columns** (left Filename and right Filename ), even when a file exists only on one side.

This enhancement would make it much easier to directly compare file names and identify missing or differing files. I hope you'll consider adding this feature in a future update.

Thank you for your hard work and dedication!

**single column Filename **
<img width="804" alt="1" src="https://github.com/user-attachments/assets/5de577de-8416-413e-9c34-3dcea84e5c15">

**sample of Both column Filename **
<img width="790" alt="2" src="https://github.com/user-attachments/assets/5f830f61-75ca-4aee-9f27-8af401f8bac9">

Issue #2535, opened 2024-11-11
does winmerge support 2 columns when do folder comparison like beyond compare?


wildwolf-test:

sdottaka:
Unfortunately, WinMerge does not support a two-column view like Beyond Compare.

guitarbug:
Maybe the two-column view can be implemented by customising the column? but there is only "Folder, Filename" instead of "Left Folder Name, Right Folder Name, Left Filename, Right Filename".

PR #3628, opened 2026-09-10
Folder compare: optional side-by-side column layout (2-way)

gongs:
## Summary
Adds an optional split-pane column layout for 2-way folder compare:
left name/size/time | status | right name/size/time, plus per-side status bar.

This is a view-only change. The compare engine and DIFFITEM model are unchanged.
3-way compare keeps the classic table.

Relates to #2535 and #2564.

## Why
WinMerge currently shows one merged table; users must read the result column
to tell left from right. File compare is already side-by-side; folder compare is not.

## How to test
1. Compare two folders with "Include subfolders".
2. View → Split Pane Layout: on/off.
3. Unique items: the missing side stays blank (no long "only on left/right" text).
4. Footer shows per-side counts and free disk space.
5. 3-way compare still uses the classic table.

## Notes
- Coding style: Allman, matching existing files.
- Default: split layout enabled for 2-way; can be turned off.
- Not in this PR: tree indent in Name columns, All/Diffs/Same display presets,
  true dual ListView, 3-pane layout.

### DIFF
--- a/Src/DirFrame.cpp
+++ b/Src/DirFrame.cpp
@@ -19,16 +19,8 @@
 #endif
 
 /**
- * @brief Statusbar pane indexes
+ * @brief Statusbar pane indexes (see CDirFrame::PANE_*)
  */
-enum
-{
-	PANE_FILTER = 1,
-	PANE_COMPMETHOD,
-	PANE_LEFT_RO,
-	PANE_MIDDLE_RO,
-	PANE_RIGHT_RO,
-};
 
 /**
  * @brief Width of compare method name pane in statusbar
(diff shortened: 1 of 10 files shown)


sdottaka:
Thank you for this PR.

Since this is a relatively large PR, I plan to review it after I finish reviewing and making fixes for PR #3515. It may take some time, possibly a month or more.

Some points I would like to consider are:

* Support for 3-way comparison.
* Adding a splitter between the two panes, with the header bars resizing/synchronizing when the splitter is moved.
* Keeping the pane proportions when the window is resized.
* Whether the center Short Result column is necessary in the 2-pane layout, since it makes pane splitting more difficult.


In addition, I am wondering whether we need a separate "Split Pane Layout" mode at all. Instead, we could consider adding separate "Left Filename" and "Right Filename" columns as optional columns, allowing users to choose which columns they want to display.

This might provide a simpler and more flexible solution while making better use of the existing ListView.

PR #3629, opened 2026-09-10
Folder compare: split-pane tree indent and All/Diffs/Same display presets

gongs:
## Summary
Follow-up to the split-pane layout PR.

When split-pane layout is on (2-way, recursive), folder compare draws tree
indent and +/- expanders in both Name columns, and adds View → Display
presets: All / Different / Identical.

Classic table + native state-image tree is unchanged when split layout is off.

Depends on: the split-pane layout PR (`gongs:feature/dirview-split-pane-layout`).
Please review the second commit if that PR is still open.

Relates to #2535 and #2564.

## How to test
1. 2-way folder compare with "Include subfolders" and Split Pane Layout on.
2. Expand/collapse from left and right Name columns; rows stay aligned.
3. View → Display → All / Different / Identical.
4. Turn Split Pane Layout off: classic table and tree mode still work.
5. 3-way compare still uses the classic table.

## Notes
- Split-pane expand-subdirs is stored separately (`DirView/SplitExpandSubdirs`).
- Tree-mode menus are disabled until compare context exists (`HasDiffs()`).


gongs:
开发前
开发后

